PR #73, opened 2019-08-11
RD-1051 - feat: add lambda-apigw-lambda support

saartochner:


src/lumigo_tracer/parsers/parser.py
```diff
@@ -78,73 +78,79 @@ def parse_response(
 
 class ServerlessAWSParser(Parser):
     def parse_response(self, url: str, status_code: int, headers, body: bytes) -> dict:
+        additional_info = {}
```
doriaviram-lumigo: Add test

PR #80, opened 2019-09-02
Rd 1102 variables

nirLumigo:


src/lumigo_tracer/utils.py
```diff
@@ -161,3 +162,62 @@ def is_aws_environment():
...
+
+    @staticmethod
+    def _truncate_locals(f_locals: Dict[str, Any]) -> FrameVariables:
+        """
+        Truncate variable part or the entire variable in order to avoid exceeding the maximum frames size.
+        :param f_locals: inspect.FrameInfo.frame.f_locals
+        """
+        locals_truncated: FrameVariables = {}
+        for var_name, var_value in f_locals.items():
+            if not callable(var_value):
```
saartochner: why?
you ignore here classes, which may be really not fun when they use the factory design pattern or something like that...

src/lumigo_tracer/utils.py
```diff
@@ -161,3 +162,62 @@ def is_aws_environment():
...
+class Frame:
+    MAX_VAR_LEN = 200
+
+    total_frames_size: int = 0
+
+    def __init__(self, frame_info: inspect.FrameInfo):
+        self.lineno: int = frame_info.lineno
+        self.file_name: str = frame_info.filename
+        self.function: str = frame_info.function
+        Frame.total_frames_size += len(str(self.lineno)) + len(self.file_name) + len(self.function)
```
saartochner: You will crash in a max_recursion_depth - because you always insert that information.
I would suggest changing the structure that you built, and create a single function (not class!) that returns the final dictionary that represents a frame, and decide outside the function whether to use it or not...

src/lumigo_tracer/utils.py
```diff
@@ -161,3 +162,62 @@ def is_aws_environment():
...
+        self.variables: FrameVariables = Frame._truncate_locals(frame_info.frame.f_locals)
```
saartochner: constructors should be simple, move this logic to `to_dict`

src/lumigo_tracer/utils.py
```diff
@@ -161,3 +162,62 @@ def is_aws_environment():
...
+
+
+def extract_frames_from_exception() -> List[Frame]:
+    """
+    Should be called only in `except` scope.
+    """
+    Frame.total_frames_size = 0
+    frames: List[Frame] = []
+    frames_infos = inspect.trace()
+    for frame_info in frames_infos[::-1]:
```
saartochner:
```suggestion
    for frame_info in frames_infos.reverse():
```
is more explicit!

src/lumigo_tracer/utils.py
```diff
@@ -161,3 +162,62 @@ def is_aws_environment():
...
+        if Frame.total_frames_size >= MAX_VARS_SIZE or "lumigo_tracer" in frame_info.filename:
+            return frames
+        frames.append(Frame(frame_info))
```
saartochner: you are reverse the order of the frames, which is not good in the case of truncated information - 
the most recent frames should be first

src/lumigo_tracer/utils.py
```diff
@@ -161,3 +162,62 @@ def is_aws_environment():
...
+    def __init__(self, frame_info: inspect.FrameInfo):
+        self.lineno: int = frame_info.lineno
+        self.file_name: str = frame_info.filename
+        self.function: str = frame_info.function
+        Frame.total_frames_size += len(str(self.lineno)) + len(self.file_name) + len(self.function)
+        self.variables: FrameVariables = Frame._truncate_locals(frame_info.frame.f_locals)
+
+    def to_dict(self) -> Dict[str, Union[int, str, FrameVariables]]:
+        return {
+            "lineno": self.lineno,
```
saartochner: change it to be non-pythonic (the rest of the tracer should have the same interface)

src/test/unit/test_main_utils.py
```diff
@@ -62,3 +75,107 @@ def test_create_request_body_take_error_first(dummy_span, error_span, function_e
...
+
+    assert frames[0].function == "func_b"
+    assert frames[0].variables == {"one": "1", "zero": "0"}
+    assert frames[1].function == "func_a"
+    assert frames[1].variables == {"a": "A"}
+    assert frames[2].function == "test_extract_frames_from_exception"
+    assert frames[2].variables == {"e": "E"}
+
+
+def test_extract_frames_from_exception_pass_max_vars_size():
```
saartochner: another test that you can add is a single parameter which is huge
`d = {i: "1" for i in range(MAX_VARS_SIZE)}`

src/test/unit/test_main_utils.py
```diff
@@ -62,3 +75,107 @@ def test_create_request_body_take_error_first(dummy_span, error_span, function_e
...
+    assert Frame._truncate_locals(f_locals) == expected
+
+
+def test_frame_truncate_locals_pass_max_vars_size():
+    f_locals = {i: "i" for i in range(MAX_VARS_SIZE * 2)}
+    actual = Frame._truncate_locals(f_locals)
+    assert len(actual) < MAX_VARS_SIZE
+
+
+def test_extract_frames_from_exception():
```
saartochner: add an explicit test for maxRecursionReached

src/lumigo_tracer/utils.py
saartochner: Ummm... I'm not sure that I'm right...
Check it :)

src/lumigo_tracer/spans_container.py
```diff
@@ -180,11 +185,15 @@ def update_event_response(
...
     def add_exception_event(self, exception: Exception) -> None:
+        """
+        Should be called only in `except` scope.
+        """
         if self.function_span:
             self.function_span["error"] = {
                 "type": exception.__class__.__name__,
                 "message": exception.args[0] if exception.args else None,
                 "stacktrace": traceback.format_exc(),
+                "frames": [frame.to_dict() for frame in extract_frames_from_exception()],
```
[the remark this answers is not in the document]
saartochner: in addition, I don't think that it makes sense to call here to two functions, and that this file (spans_container) should handle object with the `Frame` type - it should be an inner implementation of some `extract parameters`

src/lumigo_tracer/utils.py
nirLumigo: As discussed, I'm passing `free_space` so the function can decide what to do.

src/lumigo_tracer/utils.py
nirLumigo: I wanted to avoid sending function, but it doesn't really matter.

src/lumigo_tracer/utils.py
nirLumigo: `reversed(frames_infos)` :)

src/lumigo_tracer/utils.py
nirLumigo: I reverse the order of frames exactly for this purpose :)

src/lumigo_tracer/utils.py
```diff
@@ -161,3 +164,71 @@ def is_aws_environment():
     :return: heuristically determine rather we're running on an aws environment.
     """
     return bool(os.environ.get("LAMBDA_RUNTIME_DIR"))
+
+
+def extract_frames_from_exception() -> List[dict]:
```
doriaviram-lumigo: I Think its we will be  better than this function will get the `inspect` via param instead of that scopes as magic

src/lumigo_tracer/utils.py
```diff
@@ -161,3 +164,71 @@ def is_aws_environment():
...
+def convert_frame(frame_info: inspect.FrameInfo, free_space: int) -> dict:
+    return {
+        "lineno": frame_info.lineno,
+        "fileName": frame_info.filename,
+        "function": frame_info.function,
+        "variables": _truncate_locals(frame_info.frame.f_locals, free_space),
+    }
+
+
+def _truncate_locals(f_locals: Dict[str, Any], free_space: int) -> FrameVariables:
```
doriaviram-lumigo: you have the same logic at `_create_request_body` 
create a testable generic function that gets a list and size and returns pruned list

src/lumigo_tracer/utils.py
```diff
@@ -161,3 +164,71 @@ def is_aws_environment():
...
+    frames_infos = inspect.trace()
+    for frame_info in reversed(frames_infos):
+        if free_space <= 0 or "lumigo_tracer" in frame_info.filename:
+            return frames
+        frames.append(convert_frame(frame_info, free_space))
+        free_space -= len(json.dumps(frames[-1]))
+    return frames
+
+
+def convert_frame(frame_info: inspect.FrameInfo, free_space: int) -> dict:
```
doriaviram-lumigo: Add test to this functions !!!

src/lumigo_tracer/spans_container.py
```diff
@@ -180,11 +185,15 @@ def update_event_response(
...
     def add_exception_event(self, exception: Exception) -> None:
+        """
+        Should be called only in `except` scope.
+        """
         if self.function_span:
             self.function_span["error"] = {
                 "type": exception.__class__.__name__,
                 "message": exception.args[0] if exception.args else None,
                 "stacktrace": traceback.format_exc(),
+                "frames": extract_frames_from_exception() if Configuration.verbose else [],
```
doriaviram-lumigo: Get Exception trace from parm

src/test/unit/test_main_utils.py
```diff
@@ -62,3 +71,131 @@ def test_create_request_body_take_error_first(dummy_span, error_span, function_e
...
+
+    assert frames[0]["function"] == "func_b"
+    assert frames[0]["variables"] == {"one": "1", "zero": "0"}
+    assert frames[1]["function"] == "func_a"
+    assert frames[1]["variables"]["a"] == "A"
+    assert frames[2]["function"] == "test_extract_frames_from_exception"
+    assert frames[2]["variables"]["e"] == "E"
+
+
+def test_extract_frames_from_exception__max_recursion():
```
doriaviram-lumigo: Nice, I loved the tests
GJ

src/lumigo_tracer/spans_container.py
nirLumigo: The `exception` doesn't have the stack frames, I need to use `inspect` in order to go over the stack.

PR #85, opened 2019-09-19
RD-1342 - make the message always serialisable

saartochner:


src/lumigo_tracer/spans_container.py
```diff
@@ -189,9 +189,12 @@ def add_exception_event(
         self, exception: Exception, frames_infos: List[inspect.FrameInfo]
     ) -> None:
         if self.function_span:
+            message = exception.args[0] if exception.args else None
+            if not isinstance(message, str):
```
doriaviram-lumigo: Why not always?
anyway it's good :)

saartochner: readability counts!
(so the next programmer will understand that we wrote it for a reason :) )

PR #86, opened 2019-09-22
RD-1332 - feat: blacklist keys (security)

saartochner:


src/lumigo_tracer/utils.py
```diff
@@ -15,10 +16,11 @@
 LUMIGO_EVENT_KEY = "_lumigo"
 STEP_FUNCTION_UID_KEY = "step_function_uid"
 MAX_SIZE_FOR_REQUEST: int = int(os.environ.get("LUMIGO_MAX_SIZE_FOR_REQUEST", 900_000))
-MAX_VARS_SIZE = 100000
+MAX_VARS_SIZE = 100_000
 MAX_VAR_LEN = 200
 MAX_ENTRY_SIZE = 1024
 FrameVariables = Dict[str, str]
+OMITTING_KEYS_REGEXES = [".*pass.*", ".*key.*"]
```
efimk-lu: Does the user have an ability to override these?

saartochner: no.
we should give him that?
that sounds like complicated docs...

efimk-lu: I just think 
`OMITTING_KEYS_REGEXES + json.loads(os.environ.get("LUMIGO_BLACKLIST_REGEX", "[]"))`
Should not be a `+`if the env is empty then use the default

saartochner: hmm...
fair, I'll change it :)

src/lumigo_tracer/utils.py
```diff
@@ -228,3 +231,35 @@ def prepare_large_data(value: Union[str, bytes, dict, None], max_size=MAX_ENTRY_
     if len(res) > max_size:
         return f"{res[:max_size]}...[too long]"
     return res
+
+
+def get_omitting_regexes():
+    return [
+        re.compile(r, re.IGNORECASE)
+        for r in json.loads(os.environ.get("LUMIGO_BLACKLIST_REGEX", "[]")) or OMITTING_KEYS_REGEXES
```
efimk-lu: Update README

src/lumigo_tracer/utils.py
```diff
@@ -83,6 +85,7 @@ def _get_event_base64_size(event) -> int:
 def _create_request_body(
     msgs: List[dict], prune_size_flag: bool, max_size: int = MAX_SIZE_FOR_REQUEST
 ) -> str:
+    msgs = omit_keys(msgs)  # extra validation
```
efimk-lu: Just making sure, this is the omit of the entire span?

saartochner: yep

README.md
```diff
@@ -14,6 +14,12 @@ def my_lambda(event, context):
 ```
 
 ## Configuration
+* You can turn on the debug logs by setting the environment variable `LUMIGO_DEBUG=true`
+* You can change the timeout to send the trace information to the edge by setting `LUMIGO_EDGE_TIMEOUT=<seconds>`
```
efimk-lu: This is something internal, please remove it

README.md
```diff
@@ -14,6 +14,12 @@ def my_lambda(event, context):
...
+* You can prevent lumigo from sending keys that answer specific regexes by defining `LUMIGO_BLACKLIST_REGEX=["regex1", "regex2"]`. By default, we use the default regexes `[".*pass.*", ".*key.*"]`. All the regexes are case-insensitive.
+* In case of need, there is a kill switch, that stops all the interventions of lumigo immediately, without changing the code. Simply add an environment variable `LUMIGO_SWITCH_OFF=true`.
+* you can find examples in the examples directory
```
efimk-lu: Please talk with Dori, the example directory is problematic, either fix it or remove the line. See https://github.com/lumigo-io/python_tracer/pull/87

PR #90, opened 2019-10-02
RD-1400 - add domains scrubbing

saartochner:


src/lumigo_tracer/utils.py
```diff
@@ -55,6 +62,7 @@ def config(
     :param enhance_print: Should we add prefix to the print (so the logs will be in the platform).
     :param step_function: Is this function is a part of a step function?
     :param timeout_timer: Should we start a timer to send the traced data before timeout acceded.
+    :param domains_scrubber: List of regexex. We will not collect data of requests with hosts that match it.
```
nirLumigo:
```suggestion
    :param domains_scrubber: List of regexes. We will not collect data of requests with hosts that match it.
```

src/lumigo_tracer/utils.py
```diff
@@ -67,6 +75,16 @@ def config(
...
     Configuration.send_only_if_error = os.environ.get("SEND_ONLY_IF_ERROR", "").lower() == "true"
+    if "LUMIGO_DOMAINS_SCRUBBER" in os.environ:
+        try:
+            Configuration.domains_scrubber = json.loads(os.environ["LUMIGO_DOMAIN_SCRUBBER"])
+        except Exception:
+            get_logger().critical(
+                "Could not parse the specified domains scrubber, shutting down the reporter."
+            )
+            Configuration.should_report = False
+    else:
```
nirLumigo: Shouldn't it be the opposite in the if-else? Don't we prefer that the argument will override the environment variable?

src/test/unit/sync_http/test_sync_hook.py
```diff
@@ -195,6 +195,30 @@ def lambda_test_function():
...
+    def lambda_test_function():
+        return http.client.HTTPConnection(host="www.google.com").send(b"\r\n")
+
+    lambda_test_function()
+    http_events = SpansContainer.get_span().http_spans
+    assert len(http_events) == 1
+    assert "google" in http_events[0].get("info", {}).get("httpInfo", {}).get("host")
+    assert "headers" not in http_events[0]["info"]["httpInfo"]["request"]
+    assert http_events[0]["info"]["httpInfo"]["request"]["body"] == "The data is not available"
+
```
nirLumigo: Add a test with `domain_scrubber` but send a request to a website that is not supposed to be scrubbed, and make sure the span collected as before (not scrubbed)

src/lumigo_tracer/utils.py
```diff
@@ -67,6 +75,16 @@ def config(
     Configuration.is_step_function = step_function
     Configuration.timeout_timer = timeout_timer
     Configuration.send_only_if_error = os.environ.get("SEND_ONLY_IF_ERROR", "").lower() == "true"
+    if "LUMIGO_DOMAINS_SCRUBBER" in os.environ:
+        try:
+            Configuration.domains_scrubber = json.loads(os.environ["LUMIGO_DOMAIN_SCRUBBER"])
```
nirLumigo: Are you sure this is in json format?

README.md
```diff
@@ -20,6 +20,7 @@ def my_lambda(event, context):
 ## Configuration
 * You can turn on the debug logs by setting the environment variable `LUMIGO_DEBUG=true`
 * You can prevent lumigo from sending keys that answer specific regexes by defining `LUMIGO_BLACKLIST_REGEX=["regex1", "regex2"]`. By default, we use the default regexes `[".*pass.*", ".*key.*"]`. All the regexes are case-insensitive.
+* Similarly, you can prevent lumigo from sending the entire headers and body of specific domains using `LUMIGO_DOMAINS_SCRUBBER=[".*secret.*"]`, or by specify the list of regexes with the key `domains_scrubber` in the tracer's decorator. By default, we will use `["secretsmanager\..*\.amazonaws\.com", "ssm\..*\.amazonaws\.com", "kms\..*\.amazonaws\.com"]`.
```
nirLumigo: Mention that `LUMIGO_DOMAINS_SCRUBBER` is an environment variable.

src/lumigo_tracer/utils.py
saartochner: I don't think so 😮 
I think that the more flexible parameter should override less flexible, and environment variable is much more flexible than code 😮

src/lumigo_tracer/utils.py
saartochner: yep,
and that's what I demand in the readme...

do you have another idea?

src/lumigo_tracer/utils.py
saartochner: you won

PR #92, opened 2019-11-14
adding programmatic error log func

GuyMoses:


src/lumigo_tracer/utils.py
```diff
@@ -286,3 +287,8 @@ def omit_keys(value: Any):
             for k, v in value.items()
         }
     return value
+
+
+def report_error(msg: str):
+    message_with_initials = f"{LUMIGO_REPORT_ERROR_STRING} {msg}"
+    print(message_with_initials)
```
efimk-lu: Add the span_id as a prefix otherwise it won't be caught

README.md
```diff
@@ -31,6 +31,12 @@ def my_lambda(event, context):
     print('I can view this line now')
 ```
 
+### Logging Programmatic Errors
+You can use `report_error` function to write logs which will be visible in the platform.
+
+Update your import statement to `from lumigo_tracer import lumigo_tracer, report_error`.
```
efimk-lu: They might use the serverless plugin, so they won't import  `lumigo_tracer`
```suggestion
Add `from lumigo_tracer import report_error`.
```

PR #95, opened 2020-01-06
feat: specificly catch syntax error in auto instrument

saartochner:


src/lumigo_tracer/sync_http/handler.py
```diff
@@ -19,6 +19,8 @@ def _handler(*args, **kwargs):
         raise ValueError(
             "Could not find the original handler. Please follow lumigo's docs: https://docs.lumigo.io/"
         )
+    except SyntaxError:
```
efimk-lu: Now I understand, cool

src/lumigo_tracer/sync_http/sync_hook.py
```diff
@@ -7,6 +7,9 @@
 from functools import wraps
 import importlib.util
 
+# Note: This is here to increase performance - load this huge module in load time instead of runtime
```
saartochner: This is the performance improvement

PR #96, opened 2020-01-14
RD-2048 - auto instrument handle handlers paths with hierarchy

saartochner:


efimk-lu:
@saartochner on a later commit please add better logs

PR #97, opened 2020-01-15
Add support to ELB trigger by

doriaviram:


src/test/unit/parsers/test_utils.py
```diff
@@ -239,6 +239,15 @@ def test_recursive_json_join(d1, d2, result):
                 "detailType": "Scheduled Event",
             },
         ),
+        (  # elb example trigger
+            {
+                "httpMethod": "GET",
+                "resource": "resource",
+                "headers": {"Host": "www.google.com"},
+                "requestContext": {"elb": "Elb", "stage": "1", "requestId": "123"},
```
efimk-lu: That's the way it looks like, take a real example

src/test/unit/parsers/test_utils.py
```diff
@@ -239,6 +239,15 @@ def test_recursive_json_join(d1, d2, result):
...
+            },
+            {"triggeredBy": "unknown"},
```
efimk-lu: What happens when we have trigger by unknown?

src/test/unit/parsers/test_utils.py
doriaviram: 👍

src/test/unit/parsers/test_utils.py
doriaviram: image

PR #98, opened 2020-01-27
Add version to layer

efimk-lu:


scripts/bd_to_prod.sh
```diff
@@ -36,29 +36,31 @@ echo ${KEY} | gpg --batch -d --passphrase-fd 0 ${enc_location} > ~/.aws/credenti
...
-cd ../larn && npm i -g
-larn -r python3.6 -n layers/LAYERS36 --filter lumigo-python-tracer -p ~/python_tracer
-larn -r python3.7 -n layers/LAYERS37 --filter lumigo-python-tracer -p ~/python_tracer
-cd ../python_tracer
-git add layers/LAYERS36.md
-git add layers/LAYERS37.md
-git commit -m "layers-table: layers md"
+
 
 sudo pip install --upgrade bumpversion
```
saartochner: same as in node - you're uploading the package to pypi with an old version of the layers/LAYER36.md (which is ok but should be noticed)

PR #99, opened 2020-02-16
RD-2238 change environment variable name of secret masking

saartochner:


scripts/prepare_layer_files.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-pip install wheel
+pip install wheel --quiet
```
nirLumigo: Did you change that on purpose?

src/lumigo_tracer/utils.py
```diff
@@ -20,12 +20,14 @@
 MAX_VAR_LEN = 200
 MAX_ENTRY_SIZE = 1024
 FrameVariables = Dict[str, str]
-OMITTING_KEYS_REGEXES = [".*pass.*", ".*key.*"]
+OMITTING_KEYS_REGEXES = [".*pass.*", ".*key.*", ".*secret.*", ".*credential.*", ".*passphrase.*"]
```
nirLumigo: This is changing the default behavior of the tracer, it means that if someone updates the tracer the behavior will change - are we ok with that?

src/test/unit/test_main_utils.py
```diff
@@ -250,8 +254,22 @@ def test_omit_keys(value, output):
...
+
+def test_get_omitting_regexes_backward_compatibility(monkeypatch):
+    monkeypatch.setenv(LUMIGO_SECRET_MASKING_REGEX_BACKWARD_COMP, '[".*evilPlan.*"]')
+    assert [r.pattern for r in get_omitting_regexes()] == [".*evilPlan.*"]
+
+
+def test_get_omitting_regexes_fallback(monkeypatch):
+    assert [r.pattern for r in get_omitting_regexes()] == OMITTING_KEYS_REGEXES
+
+
```
nirLumigo: Add a test for a case that the user has both `LUMIGO_SECRET_MASKING_REGEX` and `LUMIGO_SECRET_MASKING_REGEX_BACKWARD_COMP`

scripts/prepare_layer_files.sh
saartochner: yes :)
It's not relevant to that PR, but just doing good in general :)

src/lumigo_tracer/utils.py
saartochner: yes,
I asked Aviad about it :)

PR #104, opened 2020-03-31
Support tags

nirLumigo:


src/lumigo_tracer/utils.py
```diff
@@ -307,3 +308,7 @@ def omit_keys(value: Any):
             for k, v in value.items()
         }
     return value
+
+
+def warn_client(msg: str) -> None:
+    print(f"{WARN_CLIENT_PREFIX}: {msg}")
```
saartochner: he should have the ability to turn off this kind of messages - do it with env_key (copy from the node tracer, it is already defined there)

src/lumigo_tracer/user_utils.py
```diff
@@ -11,3 +11,14 @@ def report_error(msg: str):
     else:
         message_with_request_id = f"RequestId: {SpansContainer.get_span().function_span.get('id')} {message_with_initials}"
         print(message_with_request_id)
+
+
+def add_tag(key: str, value: str) -> None:
```
saartochner: add flag to not print the warning in failure

src/lumigo_tracer/user_utils.py
```diff
@@ -11,3 +11,14 @@ def report_error(msg: str):
...
+    """
+    Max of 50 tags.
```
saartochner: This is customer-facing - make it trustworthy! :)

src/lumigo_tracer/spans_container.py
```diff
@@ -209,6 +212,15 @@ def add_step_end_event(self, ret_val):
             ret_val[LUMIGO_EVENT_KEY] = {STEP_FUNCTION_UID_KEY: message_id}
             get_logger().debug(f"Added key {LUMIGO_EVENT_KEY} to the user's return value")
 
+    def add_tag(self, key: str, value: str) -> bool:
+        key = str(key)
+        value = str(value)
+        tags = self.function_span["tags"]
+        if 0 < len(key) < MAX_TAG_LEN and 0 < len(value) < MAX_TAG_LEN and len(tags) < MAX_TAGS:
```
saartochner: I disagree that the value should be longer than 0 🤔

src/lumigo_tracer/spans_container.py
```diff
@@ -31,6 +31,8 @@
 MAX_BODY_SIZE = 1024
 # The buffer that we take before reaching timeout to send the traces to lumigo (seconds)
 TIMEOUT_BUFFER = 0.5
+MAX_TAGS = 50
+MAX_TAG_LEN = 50
```
saartochner: MAX_TAG_KEY_LEN = MAX_TAG_VALUE_LEN = 50

saartochner: the name of these two vars are confusing

src/lumigo_tracer/spans_container.py
```diff
@@ -87,6 +89,7 @@ def __init__(
                     "logGroupName": log_group_name,
                     **(trigger_by or {}),
                 },
+                "tags": [],
```
saartochner: note that we send it also in the `_started` span

src/test/unit/test_user_utils.py
```diff
@@ -18,3 +18,35 @@ def test_report_error_without_enhance_print(capsys):
...
+
+def test_add_tag_empty(capsys):
+    add_tag("", "value")
+    assert "Unable to add tag" in capsys.readouterr().out
+    assert SpansContainer.get_span().function_span["tags"] == []
+
+
+def test_add_tag_pass_max_chars(capsys):
+    add_tag("k" * (MAX_TAG_LEN + 1), "v" * (MAX_TAG_LEN + 1))
+    assert "Unable to add tag" in capsys.readouterr().out
```
saartochner: This is customer-facing. Print informable message!
I would move the validation to this function and print hints

src/lumigo_tracer/spans_container.py
```diff
@@ -209,6 +212,15 @@ def add_step_end_event(self, ret_val):
             ret_val[LUMIGO_EVENT_KEY] = {STEP_FUNCTION_UID_KEY: message_id}
             get_logger().debug(f"Added key {LUMIGO_EVENT_KEY} to the user's return value")
 
+    def add_tag(self, key: str, value: str) -> bool:
```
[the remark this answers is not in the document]
nirLumigo: Renamed the function in user_utils.py instead.

src/lumigo_tracer/spans_container.py
nirLumigo: Why? I think it should have value:
1. It makes sense that a name won't be empty
2. It might cause difficulties when the customer will want to search for it. (Not all DBs support empty strings)

src/lumigo_tracer/spans_container.py
nirLumigo: Ok, do you in can cause a problem?

src/lumigo_tracer/spans_container.py
saartochner: I meant specifically `value`, but it's not critical comment...
(For example, I want to tag this invocation with "important", which is a key without value...)

src/lumigo_tracer/spans_container.py
saartochner: It depends on you BE impl., so be careful :)

src/lumigo_tracer/user_utils.py
```diff
@@ -11,3 +13,38 @@ def report_error(msg: str):
     else:
         message_with_request_id = f"RequestId: {SpansContainer.get_span().function_span.get('id')} {message_with_initials}"
         print(message_with_request_id)
+
+
+def add_execution_tag(key: str, value: str, should_log_errors: bool = True) -> None:
```
saartochner: I don't think that the user knows the phrase "execution"

src/lumigo_tracer/user_utils.py
```diff
@@ -11,3 +13,38 @@ def report_error(msg: str):
...
+    """
+    The maximum number of tags is 50.
```
saartochner: Ask "Alex" to phrase something better for you.
I would change it to something like "Use this function to tag the current invocation with dynamic information, that you can query later in the platform."
Be friendly! Haha

src/lumigo_tracer/user_utils.py
```diff
@@ -11,3 +13,38 @@ def report_error(msg: str):
...
+    :param should_log_errors: Should a log message be printed in case the tag can't be added.
+    """
+    try:
+        tags_len = SpansContainer.get_span().get_tags_len()
+        key = str(key)
+        value = str(value)
+        if not key or len(key) >= MAX_TAG_KEY_LEN:
+            if should_log_errors:
+                warn_client(
+                    f"Unable to add tag: key length should be between 1 and {MAX_TAG_KEY_LEN}: {key} - {value}"
```
saartochner: add `Skipping` or something like that, so the client will know that we're not adding this key, but everything  should continue as expected

src/lumigo_tracer/user_utils.py
```diff
@@ -11,3 +13,38 @@ def report_error(msg: str):
...
         print(message_with_request_id)
+
+
+def add_execution_tag(key: str, value: str, should_log_errors: bool = True) -> None:
+    """
+    The maximum number of tags is 50.
+    :param key: Length should be between 1 and 50.
+    :param value: Length should be between 1 and 50.
+    :param should_log_errors: Should a log message be printed in case the tag can't be added.
+    """
```
saartochner: add bool return value

src/lumigo_tracer/user_utils.py
nirLumigo: The product asked to change it: we have different kinds of tags in the system so it should be `execution_id` in the platform. We want the same language in the tracer and the platform.

src/lumigo_tracer/user_utils.py
nirLumigo: You're right, at first I didn't understand exactly what you meant :)

efimk-lu:
@saartochner pay attention 
```
<testsuite name="Test compatibilities JVM languages" timestamp="2020-04-02T07:58:03" tests="3" file="/root/integration-tests/src/test/test0903_compatiblities.js" time="270.9500" failures="1">
    <testcase name=" Test compatibilities JVM languages complex chalice environment" time="270.9500" classname="complex chalice environment">
```

The chalice test fails

src/lumigo_tracer/utils.py
```diff
@@ -10,6 +10,7 @@
 from base64 import b64encode
 import inspect
 
+EXECUTION_TAGS_KEY = "lumigo_execution_tags_no_scrub"
```
saartochner, line 13: I wouldn't add the `_no_scrub`...

src/lumigo_tracer/utils.py
```diff
@@ -38,6 +39,7 @@
     r"kms\..*\.amazonaws\.com",
     r"sts\..*amazonaws\.com",
 ]
+SKIP_SCRUBBING_KEYS = [EXECUTION_TAGS_KEY]
```
saartochner, line 42: I think that it better be configurable from the user (e.g. by envs), wdyt?

PR #106, opened 2020-04-22
remove none value data from api gw event

uri-p:


src/lumigo_tracer/spans_container.py
```diff
@@ -265,7 +266,7 @@ def create_span(cls, event=None, context=None, force=False) -> "SpansContainer":
         if Configuration.verbose:
             additional_info.update(
                 {
-                    "event": prepare_large_data(omit_keys(event)),
+                    "event": prepare_large_data(omit_keys(parse_event(event))),
```
saartochner, line 269: why is it here?
I think that it better be in the `parser` file (which is the file that parses the HTTP request and response).
The infra there already support the structure of finding the "right" praser, and parse the data.

src/lumigo_tracer/parsers/event_parser.py
```diff
@@ -0,0 +1,75 @@
...
+            return event
+    except Exception:
+        return event
+
+
+def _is_api_gw_event(event: Dict) -> bool:
+    return API_GW_REGEX.match(event["requestContext"]["domainName"]) is not None
+
+
+def _parse_api_gw_event(event: Dict):
```
saartochner, line 47: As I wrote in the concluence, this function should be dependent on some user input (env var?) to decide whether to cut the body or not....

src/lumigo_tracer/parsers/event_parser.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+def parse_event(event: Dict):
+    try:
+        if _is_api_gw_event(event):
+            return _parse_api_gw_event(event)
+        else:
+            return event
+    except Exception:
+        return event
```
saartochner, line 40: add debug message

src/lumigo_tracer/parsers/event_parser.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+def _is_api_gw_event(event: Dict) -> bool:
```
saartochner, line 43: what if they're behind DNS? (like Toyota 😂 )
We can use the HTTP headers to decide !

src/lumigo_tracer/parsers/event_parser.py
```diff
@@ -0,0 +1,75 @@
...
+    return API_GW_REGEX.match(event["requestContext"]["domainName"]) is not None
+
+
+def _parse_api_gw_event(event: Dict):
+    new_event: OrderedDict = OrderedDict()
+    # Add order keys
+    for order_key in API_GW_KEYS_ORDER:
```
saartochner, line 50: are you sure that there aren't any more keys?
We're going to omit all the other keys - maybe we want to keep them *first*?

src/lumigo_tracer/parsers/event_parser.py
uri-p: Can you give an example ?

src/lumigo_tracer/parsers/event_parser.py
```diff
@@ -0,0 +1,75 @@
...
+        if event.get(order_key):
+            new_event[order_key] = event[order_key]
```
saartochner, line 52:
```suggestion
            new_event = {k: v for k, v in event.items() if k in API_GW_KEYS_ORDER}
```
Is more pythonic :)

src/lumigo_tracer/parsers/event_parser.py
uri-p: This is based on the AWS schema of API gateway.
All other keys are added in the end

src/lumigo_tracer/parsers/event_parser.py
```diff
@@ -0,0 +1,75 @@
+import re
+from collections import OrderedDict
+from typing import Dict
+
+API_GW_REGEX = re.compile(rf".*execute-api.*amazonaws\.com.*")
```
saartochner, line 5:
```suggestion
API_GW_REGEX = re.compile(r".*execute-api.*amazonaws\.com.*")
```

saartochner: I think 🤔

src/lumigo_tracer/parsers/utils.py
```diff
@@ -274,3 +274,15 @@ def should_scrub_domain(url: str) -> bool:
             if re.match(regex, url, re.IGNORECASE):
                 return True
     return False
+
+
+def str_to_list(val: str):
+    if val:
```
saartochner, line 280: add try/except!
It runs in global time, and if there will be error- then it will crush the customer's lambda

src/lumigo_tracer/parsers/event_parser.py
```diff
@@ -0,0 +1,93 @@
...
+) or ("cookie", "X-Amz", "Accept", "CloudFront", "Via", "X-Forwarded", "sec-")
+API_GW_REQUEST_CONTEXT_FILTER_KEYS = str_to_list(
+    os.environ.get("LUMIGO_API_GW_REQUEST_CONTEXT_FILTER_KEYS", "")
+) or ["authorizer"]
+API_GW_KEYS_DELETE_KEYS = str_to_list(os.environ.get("LUMIGO_API_GW_KEYS_DELETE_KEYS", "")) or [
+    "multiValueHeaders"
+]
+
+
+class EventParseHandler:
```
saartochner, line 33: inherit from ABC (like the alert repository) to enforce implementation

src/lumigo_tracer/parsers/event_parser.py
```diff
@@ -0,0 +1,93 @@
...
+    def is_supported(self, event) -> bool:
+        raise NotImplementedError()
+
+    def parse(self, event):
```
saartochner, line 37: what does it return?

src/lumigo_tracer/parsers/event_parser.py
```diff
@@ -0,0 +1,93 @@
...
+        raise NotImplementedError()
+
+
+class ApiGWHandler(EventParseHandler):
+    def is_supported(self, event) -> bool:
+        if event.get("requestContext") and event.get("requestContext").get("domainName"):
```
saartochner, line 43:
```suggestion
        if event.get("requestContext") and event.get("requestContext", {}).get("domainName"):
```
So if it doesn't exists, then it will not crush on NoneType...
add test!

src/lumigo_tracer/parsers/event_parser.py
```diff
@@ -0,0 +1,93 @@
...
+        new_event: OrderedDict = OrderedDict()
+        # Add order keys
+        for order_key in API_GW_KEYS_ORDER:
+            if event.get(order_key):
+                new_event[order_key] = event[order_key]
+        # Remove requestContext keys
+        if new_event.get("requestContext"):
+            delete_request_context_keys = [
+                x
+                for x in new_event["requestContext"].keys()
```
saartochner, line 57:
```suggestion
                for x in new_event["requestContext"]
```

src/lumigo_tracer/parsers/event_parser.py
```diff
@@ -0,0 +1,93 @@
...
+                if x not in API_GW_REQUEST_CONTEXT_FILTER_KEYS
+            ]
+            for delete_request_context_key in delete_request_context_keys:
```
saartochner, line 60: why do you do it in two stages? why not as part of the for?

src/lumigo_tracer/parsers/event_parser.py
```diff
@@ -0,0 +1,93 @@
...
+class EventParseHandler:
+    def is_supported(self, event) -> bool:
+        raise NotImplementedError()
+
+    def parse(self, event):
+        raise NotImplementedError()
+
+
+class ApiGWHandler(EventParseHandler):
+    def is_supported(self, event) -> bool:
```
saartochner, line 42: make it a static function

src/lumigo_tracer/parsers/event_parser.py
```diff
@@ -0,0 +1,93 @@
...
+            for delete_headers_key in delete_headers_keys:
+                new_event["headers"].pop(delete_headers_key, None)
+        # Add all other keys
+        for key in event.keys():
+            if (key not in API_GW_KEYS_ORDER) and (key not in API_GW_KEYS_DELETE_KEYS):
+                new_event[key] = event[key]
+        return new_event
+
+
+HANDLERS: List[EventParseHandler] = [ApiGWHandler()]
```
saartochner, line 78: Don't initiate objects here, it is more correct to use static methods

src/lumigo_tracer/parsers/event_parser.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+class EventParser:
+    @staticmethod
+    def parse_event(event: Dict, handlers: List[EventParseHandler] = HANDLERS):
```
saartochner, line 83: never use mutable objects as defaults in signatures - isn't pycharm marked it for you?
change it to None, and in the function do something like
`handlers = handlers or DEFAULT_HANDLERS`

src/test/unit/parsers/test_utils.py
```diff
@@ -310,3 +310,19 @@ def test_safe_get(d, keys, result_value, default):
...
     Configuration.domains_scrubber = regexes
     assert should_scrub_domain(url) == expected
+
+
+def test_str_to_list():
+    assert str_to_list("a,b,c,d") == ["a", "b", "c", "d"]
+
+
+def test_str_to_list_exception():
+    assert str_to_list(None) is None
```
saartochner, line 320: give it also an empty list (which is your actual default)

src/lumigo_tracer/parsers/event_parser.py
uri-p: I got an exception of iterator change

saartochner: use `.copy()`,
it is more readable :)

uri-p: But it will affect my order dict no ?

saartochner:
```suggestion
            for delete_request_context_key in new_event["requestContext"].copy():
```
shouldn't, but It's not that important.....

src/lumigo_tracer/parsers/event_parser.py
```diff
@@ -0,0 +1,114 @@
...
+        if new_event.get("headers"):
+            delete_headers_keys = [
+                x
+                for x in new_event["headers"]
+                if x.startswith(API_GW_PREFIX_KEYS_HEADERS_DELETE_KEYS)
+            ]
+            for delete_headers_key in delete_headers_keys:
+                new_event["headers"].pop(delete_headers_key, None)
+        # Add all other keys
+        for key in event.keys():
```
saartochner, line 95: Are you sure that you want to put the other keys in the end?

src/lumigo_tracer/parsers/event_parser.py
```diff
@@ -0,0 +1,114 @@
...
+    @staticmethod
+    def parse_event(event: Dict, handlers: List[EventParseHandler] = None):
+        handlers = handlers or [ApiGWHandler()]
+        for handler in handlers:
+            try:
+                if handler.is_supported(event):
+                    return handler.parse(event)
+            except Exception as e:
+                get_logger().debug(
+                    f"Error while trying to parse with handler {type(handler)} event {event}",
```
saartochner, line 111:
```suggestion
                    f"Error while trying to parse with handler {handler.__class__.__name__} event {event}",
```

src/lumigo_tracer/parsers/utils.py
```diff
@@ -274,3 +274,21 @@ def should_scrub_domain(url: str) -> bool:
...
                 return True
     return False
+
+
+def str_to_list(val: str):
+    try:
+        if val:
+            return val.split(",")
+    except Exception:
+        pass
```
saartochner, line 284: log?

src/test/unit/parsers/test_utils.py
```diff
@@ -310,3 +310,19 @@ def test_safe_get(d, keys, result_value, default):
...
     Configuration.domains_scrubber = regexes
     assert should_scrub_domain(url) == expected
+
+
+def test_str_to_list():
+    assert str_to_list("a,b,c,d") == ["a", "b", "c", "d"]
+
+
+def test_str_to_list_exception():
+    assert str_to_list([]) is None
```
saartochner, line 320:
```suggestion
    assert str_to_list("") is None
```
I meant that (because that's how you call it when you define your variables...)

PR #109, opened 2020-04-27
Add missing header filter

uri-p:


src/test/unit/parsers/test_event_parser.py
```diff
@@ -15,6 +15,12 @@ def parse(event) -> Dict:
         raise Exception()
 
 
+def test_parse_event_not_api_gw_none_check():
+    new_event = EventParser.parse_event(event=None, handlers=[ExceptionHandler()])
```
saartochner, line 19: event can not be None... 🤔

saartochner: why did you changed it?

PR #114, opened 2020-05-31
timeout buffer become configurable

saartochner:


src/lumigo_tracer/utils.py
```diff
@@ -96,6 +100,9 @@ def config(
         step_function or os.environ.get("LUMIGO_STEP_FUNCTION", "").lower() == "true"
     )
     Configuration.timeout_timer = timeout_timer
+    Configuration.timeout_timer_buffer = timeout_timer_buffer or float(
```
efimk-lu, line 103: What happens if someone will write there a string or space by accident?

saartochner: it will not work, such like every other configuration that the customer fills mis-correctly... 🤔

saartochner: The client needs to implicitly write `timeout_timer_buffer='  '` in order for it to not work...
(empty string and none will work)

efimk-lu: I mean, will the tracer behave correctly in case it tries to parse a non float value

saartochner: It will raise an exception immediately

efimk-lu: That's the problem I was referring to. Bad configuration, just load the default and write to logs print(bla bla) don't put this error message under lumigo debug

saartochner: Updated

PR #117, opened 2020-06-29
RD-3149 - handle large downloads

saartochner:
IT will follow!

### DIFF
--- a/src/lumigo_tracer/spans_container.py
+++ b/src/lumigo_tracer/spans_container.py
@@ -16,6 +16,7 @@ from lumigo_tracer.utils import (
     prepare_large_data,
     omit_keys,
     EXECUTION_TAGS_KEY,
+    MAX_ENTRY_SIZE,
 )
 from lumigo_tracer import utils
 from lumigo_tracer.parsers.parser import get_parser, HTTP_TYPE, StepFunctionParser
@@ -182,7 +183,8 @@ class SpansContainer:
                 self.previous_response_body = b""
 
             parser = get_parser(host, headers)()  # type: ignore
-            self.previous_response_body += body
+            if len(self.previous_response_body) < MAX_ENTRY_SIZE:
+                self.previous_response_body += body
             update = parser.parse_response(  # type: ignore
                 host, status_code, headers, self.previous_response_body
             )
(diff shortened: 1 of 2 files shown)


src/test/unit/test_spans_container.py
```diff
@@ -156,3 +156,19 @@ def test_get_tags_len():
     SpansContainer.get_span().add_tag("k0", "v0")
     SpansContainer.get_span().add_tag("k1", "v1")
     assert SpansContainer.get_span().get_tags_len() == 2
+
+
+def test_aggregating_response_body(dummy_http_request):
+    """
+    This test is herre to validate that we're not leaking memory on aggregating response body.
```
nirLumigo, line 163:
```suggestion
    This test is here to validate that we're not leaking memory on aggregating response body.
```

src/test/unit/test_spans_container.py
```diff
@@ -156,3 +156,19 @@ def test_get_tags_len():
...
+    Unfortunately python doesn't give us better tools, so we most check the problematic member itself.
```
nirLumigo, line 164:
```suggestion
    Unfortunately python doesn't give us better tools, so we must check the problematic member itself.
```

src/test/unit/test_spans_container.py
```diff
@@ -156,3 +156,19 @@ def test_get_tags_len():
...
+    """
+    SpansContainer.create_span()
+    SpansContainer.get_span().add_request_event(dummy_http_request)
+
+    big_response_chunk = b"leak" * MAX_ENTRY_SIZE
+    for _ in range(10):
```
nirLumigo, line 170: Why do you do it 10 times?

src/lumigo_tracer/spans_container.py
```diff
@@ -182,7 +183,8 @@ def update_event_response(
                 self.previous_response_body = b""
 
             parser = get_parser(host, headers)()  # type: ignore
-            self.previous_response_body += body
+            if len(self.previous_response_body) < MAX_ENTRY_SIZE:
```
nirLumigo, line 186: Don't we take only the last 1024 bytes?

PR #118, opened 2020-07-05
Rd-3162 - return same value

saartochner:


src/lumigo_tracer/utils.py
```diff
@@ -273,13 +275,15 @@ def prepare_large_data(
 
     :param value: The value we wish to send
     :param max_size: The maximum size of the data that we will send
+    :param enforce_jsonify: Should we accept exception in the jsonify
```
nirLumigo, line 278:
```suggestion
    :param enforce_jsonify: Should we raise exception in the jsonify
```

src/lumigo_tracer/utils.py
```diff
@@ -262,7 +262,9 @@ def _truncate_locals(f_locals: Dict[str, Any], free_space: int) -> FrameVariable
 
 
 def prepare_large_data(
-    value: Union[str, bytes, dict, OrderedDict, None], max_size=MAX_ENTRY_SIZE
+    value: Union[str, bytes, dict, OrderedDict, None],
+    max_size=MAX_ENTRY_SIZE,
+    enforce_jsonify=False,
```
nirLumigo, line 267:
```suggestion
    enforce_jsonify: bool=False,
```

src/test/unit/sync_http/test_sync_hook.py
```diff
@@ -532,6 +532,21 @@ def lambda_test_function(event, context):
     ]
 
 
+def test_not_jsonable_return(monkeypatch):
+    @lumigo_tracer()
+    def lambda_test_function(event, context):
+        return {"a": open("/tmp/a.txt", "wb")}
```
nirLumigo, line 538: If the test will run on the machine without `"/tmp"` the test will fail.
You can use another object that isn't json serializable and isn't dependant on the machine running the date, e.g `datetime.datetime`.

PR #119, opened 2020-07-06
RD-3150 - code - optimization of the reporting process

saartochner:
On lambda with 256MB that executes 3000 ddb `get_item` requests, reducing the reporting time by 58%. 
Moreover, this puts an upper bound to the duration of the final report, which decreases as the memory of the lambda increases.

### DIFF
--- a/src/lumigo_tracer/utils.py
+++ b/src/lumigo_tracer/utils.py
@@ -17,9 +17,8 @@ LOG_FORMAT = "#LUMIGO# - %(asctime)s - %(levelname)s - %(message)s"
 SECONDS_TO_TIMEOUT = 0.5
 LUMIGO_EVENT_KEY = "_lumigo"
 STEP_FUNCTION_UID_KEY = "step_function_uid"
-TOO_BIG_SPANS_THRESHOLD = (
-    5
-)  # number of spans that are too big to enter the reported message before break
+# number of spans that are too big to enter the reported message before break
+TOO_BIG_SPANS_THRESHOLD = 5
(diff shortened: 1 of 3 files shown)


src/lumigo_tracer/utils.py
```diff
@@ -280,7 +297,7 @@ def prepare_large_data(
     """
     if isinstance(value, dict) or isinstance(value, OrderedDict):
         try:
-            value = json.dumps(value)
+            value = json.dumps(value, cls=DecimalEncoder)
```
saartochner, line 300: This is not optimization but fixing a bug

src/lumigo_tracer/utils.py
```diff
@@ -138,22 +141,30 @@ def _create_request_body(
...
     max_size: int = MAX_SIZE_FOR_REQUEST,
-    min_size: int = MIN_SPAN_SIZE,
+    too_big_spans_threshold: int = TOO_BIG_SPANS_THRESHOLD,
 ) -> str:
-    if not prune_size_flag or _get_event_base64_size(msgs) < max_size:
+
+    if not prune_size_flag or (
+        len(msgs) < NUMBER_OF_SPANS_IN_REPORT_OPTIMIZATION
+        and _get_event_base64_size(msgs) < max_size  # noqa
+    ):
```
doriaviram, line 150: Running only 1 time hereof `_get_event_base64_size(msgs)` is effecting so much ?

saartochner: hell yeah!
this depends on the number of spans. In my tests (3000 spans) it took a full second!

PR #121, opened 2020-07-08
RD-3163 - ddb streams

saartochner:


src/lumigo_tracer/parsers/parser.py
```diff
@@ -94,13 +95,39 @@ def parse_response(self, url: str, status_code: int, headers, body: bytes) -> di
 class DynamoParser(ServerlessAWSParser):
     should_add_message_id = False
 
+    def _extract_message_id(self, body: dict, method: str):
```
nirLumigo, line 98: Shouldn't it be a staticmethod?

nirLumigo, line 98:
```suggestion
    def _extract_message_id(self, body: dict, method: str) -> returnType:
```

src/lumigo_tracer/parsers/parser.py
```diff
@@ -94,13 +95,39 @@ def parse_response(self, url: str, status_code: int, headers, body: bytes) -> di
...
+            return md5hash(body["Item"])
+        elif method in ("UpdateItem", "DeleteItem") and body.get("Key"):
+            return md5hash(body["Key"])
+        elif method == "BatchWriteItem" and body.get("RequestItems"):
+            first_item = next(iter(body["RequestItems"].values()))
+            if first_item:
+                if first_item[0].get("PutRequest"):
+                    return md5hash(first_item[0]["PutRequest"]["Item"])
+                else:
+                    return md5hash(first_item[0]["DeleteRequest"]["Key"])
```
nirLumigo, line 109: Are we sure it will have `"DeleteRequest"` and `"Key"`? If not it may raise an exception

src/lumigo_tracer/parsers/parser.py
```diff
@@ -94,13 +95,39 @@ def parse_response(self, url: str, status_code: int, headers, body: bytes) -> di
...
+    def _extract_message_id(self, body: dict, method: str):
+        if method == "PutItem" and body.get("Item"):
+            return md5hash(body["Item"])
+        elif method in ("UpdateItem", "DeleteItem") and body.get("Key"):
+            return md5hash(body["Key"])
+        elif method == "BatchWriteItem" and body.get("RequestItems"):
+            first_item = next(iter(body["RequestItems"].values()))
+            if first_item:
+                if first_item[0].get("PutRequest"):
+                    return md5hash(first_item[0]["PutRequest"]["Item"])
```
nirLumigo, line 107: Are we sure it will have `"Item"`? If not, it may raise an exception.

src/lumigo_tracer/parsers/parser.py
```diff
@@ -94,13 +95,39 @@ def parse_response(self, url: str, status_code: int, headers, body: bytes) -> di
...
+                else:
+                    return md5hash(first_item[0]["DeleteRequest"]["Key"])
+
+    def _extract_table_name(self, body: dict, method: str):
+        name = body.get("TableName")
+        if not name and method == "BatchWriteItem" and isinstance(body.get("RequestItems"), dict):
+            return next(iter(body["RequestItems"]))
```
nirLumigo, line 114: If `body["RequestItems"] == {}` (empty dict) you'll get an exception:
image

src/lumigo_tracer/parsers/parser.py
```diff
@@ -94,13 +95,39 @@ def parse_response(self, url: str, status_code: int, headers, body: bytes) -> di
...
+        return name
+
     def parse_request(self, parse_params: HttpRequest) -> dict:
         target: str = parse_params.headers.get("x-amz-target", "")
+        method = safe_split_get(target, ".", 1)
+        try:
+            parsed_body = json.loads(parse_params.body)
+        except json.JSONDecodeError:
```
nirLumigo, line 122: Log the error

src/lumigo_tracer/parsers/utils.py
```diff
@@ -289,9 +290,25 @@ def _parse_streams(event: dict) -> Dict[str, str]:
         result["messageId"] = event["Records"][0].get("messageId")
     elif triggered_by == "kinesis":
         result["messageId"] = safe_get(event, ["Records", 0, "kinesis", "sequenceNumber"])
+    elif triggered_by == "dynamodb":
+        result.update(_parse_dynamomdb_event(event))
     return result
 
 
+def _parse_dynamomdb_event(event):
```
nirLumigo, line 298:
```suggestion
def _parse_dynamomdb_event(event) -> returnType:
```

src/lumigo_tracer/parsers/utils.py
```diff
@@ -289,9 +290,25 @@ def _parse_streams(event: dict) -> Dict[str, str]:
...
+    creation_time = (
+        event["Records"][0].get("dynamodb", {}).get("ApproximateCreationDateTime", 0) * 1000
```
nirLumigo, line 300: Why ` * 1000`? Maybe extract it to a function so it will be more clear.

src/lumigo_tracer/parsers/utils.py
```diff
@@ -289,9 +290,25 @@ def _parse_streams(event: dict) -> Dict[str, str]:
...
+    )
+    mids = []
+    for record in event["Records"]:
+        event_name = record.get("eventName")
+        if event_name in ("MODIFY", "REMOVE") and record.get("dynamodb", {}).get("Keys"):
+            mids.append(md5hash(record["dynamodb"]["Keys"]))
+        elif event_name == "INSERT" and record.get("dynamodb", {}).get("NewImage"):
+            mids.append(md5hash(record["dynamodb"]["NewImage"]))
+    return {"messageIds": mids, "approxEventCreationTime": creation_time}
```
nirLumigo, line 309: IMO it will be nice to replace the keys with constants.

src/lumigo_tracer/utils.py
```diff
@@ -383,3 +384,9 @@ def get_timeout_buffer(remaining_time: float):
     if not buffer:
         buffer = max(0.5, min(0.1 * remaining_time, 3))
     return buffer
+
+
+def md5hash(d: dict):
```
nirLumigo, line 389:
```suggestion
def md5hash(d: dict) -> returnType:
```

src/test/unit/parsers/test_parser.py
```diff
@@ -61,3 +72,68 @@ def test_apigw_parse_response_with_aws_request_id():
...
+        host="",
+        method="POST",
+        uri="",
+        headers={"x-amz-target": "DynamoDB_20120810.GetItem"},
+        body="not a json",
+    )
+    response = parser.parse_request(params)
+    assert response["info"]["resourceName"] is None
+    assert response["info"]["dynamodbMethod"] == "GetItem"
+    assert response["info"]["messageId"] is None
```
nirLumigo, line 139: Maybe add a test in case we get a `"dynamodbMethod"` that we don't support.

src/lumigo_tracer/utils.py
```diff
@@ -383,3 +384,9 @@ def get_timeout_buffer(remaining_time: float):
...
+    h = hashlib.md5()
+    h.update(json.dumps(d, sort_keys=True).encode())
```
nirLumigo, line 391: 💪

src/lumigo_tracer/parsers/parser.py
saartochner: This is not a legal query - boto3 will return an error...

src/lumigo_tracer/parsers/parser.py
saartochner: If, otherwise boto3 will crash

src/lumigo_tracer/parsers/parser.py
saartochner: If, otherwise boto3 will crash

src/lumigo_tracer/parsers/utils.py
saartochner: seconds to milliseconds..
I will 👍

src/test/unit/parsers/test_parser.py
```diff
@@ -137,3 +137,16 @@ def test_dynamodb_parser_sad_flow():
...
+def test_dynamodb_parser_sad_flow_unsupported_query():
+    parser = DynamoParser()
+    params = HttpRequest(
+        host="",
+        method="POST",
+        uri="",
+        headers={"x-amz-target": "DynamoDB_20120810.BatchWriteItem"},
+        body='{"RequestItems": {}}',
+    )
+    with pytest.raises(Exception):
```
nirLumigo, line 151: In case we don't support the query, we don't want to throw an exception, we just won't collect the data.

PR #123, opened 2020-07-19
RD-3180 - feat: scrubbing and dump json simultaneously

saartochner:


src/lumigo_tracer/libs/custom_json_dump.py
```diff
@@ -0,0 +1,80 @@
+from json.encoder import (
```
saartochner, line 1: Note this is a lib file that was copied from StackOverflow (with small changes).
We're using it in order to not have any external libraries.

PR #124, opened 2020-07-19
RD-3180 - scrubbing big spans

saartochner:


src/lumigo_tracer/libs/custom_json_dump.py
```diff
@@ -0,0 +1,80 @@
+from json.encoder import (
```
saartochner, line 1: Note this is a lib file that was copied from StackOverflow (with small changes).
We're using it in order to not have any external libraries.

src/lumigo_tracer/utils.py
```diff
@@ -259,68 +266,32 @@ def format_frame(frame_info: inspect.FrameInfo, free_space: int) -> dict:
...
-        var = {var_name: prepare_large_data(var_value, MAX_VAR_LEN)}
+        if any(r.match(var_name) for r in omitting_regexes):  # type: ignore
+            var = {var_name: "****"}
+        else:
+            var = {var_name: lumigo_dumps(var_value, max_size=MAX_VAR_LEN)}
         free_space -= len(json.dumps(var))
         if free_space <= 0:
             return locals_truncated
         locals_truncated.update(var)
     return locals_truncated
```
doriaviram, line 292: Seems like you can migrate it to use `lumigo_dumps`, you have here also scrub and json.dump
It's better to have In encapsulated service like your `lumigo_dumps`

src/test/unit/parsers/test_parser.py
```diff
@@ -45,7 +45,7 @@ def test_apigw_parse_response():
             "host": "dummy",
             "response": {
                 "headers": '{"apigw-requestid": "LY_66j0dPHcESCg="}',
-                "body": "",
+                "body": '""',
```
doriaviram, line 48: Think that empty should stay `""` and not `'""'`, Better to send `null` I think if there is no-body, but second choice is `""`

src/test/unit/test_main_utils.py
```diff
@@ -194,27 +197,57 @@ def func():
...
-        # dict that can't be converted to json.
-        ({"a": set()}, "{'a': set()}"),  # type: ignore
-        (b"a", "a"),  # bytes that can be decoded.
-        (b"\xff\xfea\x00", "b'\\xff\\xfea\\x00'"),  # bytes that can't be decoded.
-        ({1: Decimal(1)}, '{"1": 1.0}'),  # decimal should be serializeable
+        ("aa", '"aa"'),  # happy flow - string
+        (None, "null"),  # happy flow - None
+        ("a" * 101, '"' + "a" * 99 + "...[too long]"),  # simple long string
+        ({"a": "a"}, '{"a": "a"}'),  # happy flow - dict
+        ({"a": set()}, '{"a": "set()"}'),  # dict that can't be converted to json
```
doriaviram, line 212: ?
Think you need to solve this

src/lumigo_tracer/parsers/event_parser.py
```diff
@@ -69,7 +70,7 @@ def parse(event) -> Dict:
         # Add order keys
         for order_key in API_GW_KEYS_ORDER:
             if event.get(order_key):
-                new_event[order_key] = event[order_key]
+                new_event[order_key] = copy.deepcopy(event[order_key])
```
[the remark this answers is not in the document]
saartochner: to not change the event.
(for example, We later use it to deduce messgaeId)

src/lumigo_tracer/spans_container.py
```diff
@@ -286,12 +287,16 @@ def create_span(cls, event=None, context=None, force=False) -> "SpansContainer":
         """
         if cls._span and not force:
             return cls._span
+        # copy the event to ensure that we will not change it
+        event = copy.deepcopy(event)
```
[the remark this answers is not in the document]
saartochner: it's on purpise - 
This one is here to make sure that we will not change the parameters for the customers.
The one on the parser is because this specific parser changed the value, and we want to continue using the original one...

src/lumigo_tracer/utils.py
```diff
@@ -152,7 +159,7 @@ def _create_request_body(
         len(msgs) < NUMBER_OF_SPANS_IN_REPORT_OPTIMIZATION
         and _get_event_base64_size(msgs) < max_size  # noqa
     ):
-        return json.dumps(omit_keys(msgs))
+        return lumigo_dumps(msgs, max_size=max_size)
```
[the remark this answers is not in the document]
saartochner: yes... We always did this twice to ensure that we're not leaking data.
Does the node tracer doesn't do it? do you think that we better remove it?

src/test/unit/parsers/test_parser.py
saartochner: Note that `""` is not a valid json - it's actually nothing (the quotes are pythonics).
There is either:
1. An empty string, which is is `""` (so in python it will look like `'""'`)
2. Or null, which is `null` (so in python it will look like `"null"`)
And they represent different values...

src/test/unit/test_main_utils.py
saartochner: that's on purpose 😂 
what should we do if the object is not serializable?

src/lumigo_tracer/utils.py
saartochner: I wish :(
I don't have json.dump here :(

src/lumigo_tracer/utils.py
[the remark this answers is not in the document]
doriaviram: Yes, Create a tests to validate the "full flow"

src/lumigo_tracer/utils.py
doriaviram: What do you mean? `free_space -= len(json.dumps(var))`
Anyway, don't repeat here the omit logic & dump, use the dump func

src/lumigo_tracer/spans_container.py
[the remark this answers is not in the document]
doriaviram: Look like the Parser should be pure.. NVM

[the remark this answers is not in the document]
saartochner: agree...
This makes it pure :)

src/lumigo_tracer/utils.py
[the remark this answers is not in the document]
saartochner: what do you mean by creating a test?
The idea of this line is that something which is not predictable will leak here... so what do you think that the test should check? 😅

src/lumigo_tracer/utils.py
saartochner: this function only dumps to calculate size - it does not really dump...
This is the problem with merging the omit and the dump - Any function that do only one thing (only omit in this case) could not use it...

PR #126, opened 2020-07-26
RD 3180 scrubbing big spans

saartochner:


src/lumigo_tracer/parsers/utils.py
```diff
@@ -147,7 +147,7 @@ def recursive_json_join(d1: dict, d2: dict):
     for key in set(itertools.chain(d1.keys(), d2.keys())):
         value = d1.get(key, d2.get(key))
         if isinstance(value, dict):
-            d[key] = recursive_json_join(d1.get(key, {}), d2.get(key, {}))
```
saartochner, line 150: trim the non-existing branch fast

src/lumigo_tracer/sync_http/sync_hook.py
```diff
@@ -56,7 +56,7 @@ def _request_wrapper(func, instance, args, kwargs):
             hooked_headers = getattr(instance, LUMIGO_HEADERS_HOOK_KEY, None)
             if hooked_headers:
                 # we will get here only if _headers_reminder_wrapper ran first. remove its traces.
-                headers = {ensure_str(k): ensure_str(v) for k, v in hooked_headers.items()}
```
saartochner, line 59: unnecessary (and saves 1s / 2500 requests)

src/lumigo_tracer/utils.py
```diff
@@ -165,7 +172,7 @@ def _create_request_body(
         len(msgs) < NUMBER_OF_SPANS_IN_REPORT_OPTIMIZATION
         and _get_event_base64_size(msgs) < max_size  # noqa
     ):
-        return json.dumps(omit_keys(msgs))
```
saartochner, line 168: we already omitted the keys

src/lumigo_tracer/utils.py
```diff
@@ -322,83 +329,17 @@ def default(self, o):
...
-                items[k] = v
-            elif isinstance(k, str) and any(r.match(k) for r in regexes):  # type: ignore
-                items[k] = "****"
-            else:
-                items[k] = omit_keys(v, regexes)
-        return items
-    return value
+    if not given_regexes:
+        return ""
+    return re.compile(fr"({'|'.join(given_regexes)})", re.IGNORECASE)
```
saartochner, line 342: create a single regex Pattern that holds all the options (lowering the `any` to c-level)

src/test/unit/test_main_utils.py
```diff
@@ -234,60 +268,29 @@ def test_format_frame():
...
-        ('{"hello": "world", "password": "abc"}', '{"hello": "world", "password": "****"}'),
-        (b'{"hello": "world", "password": "abc"}', '{"hello": "world", "password": "****"}'),
-        ('{"hello": "w', '{"hello": "w'),
-        ("5", "5"),
-        ([{"password": 1}, {"a": "b"}], [{"password": "****"}, {"a": "b"}]),
-        ({None: 1}, {None: 1}),
-        ({SKIP_SCRUBBING_KEYS[0]: {"password": 1}}, {SKIP_SCRUBBING_KEYS[0]: {"password": 1}}),
-    ),
-)
-def test_omit_keys(value, output):
```
saartochner, line 263: merged to the `test_lumigo_dumps` parametrize test

.bumpversion.cfg
```diff
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.1.103
```
saartochner, line 2: manually change that in order to avoid conflict when merging now

scripts/bd_to_prod.sh
```diff
@@ -41,15 +41,14 @@ echo "Getting latest changes from git"
 changes=$(git log $(git describe --tags --abbrev=0)..HEAD --oneline)
 
 
-
-sudo pip install --upgrade bumpversion
-bumpversion patch --message "{current_version} → {new_version}. Changes: ${changes}"
-
-
 echo "Uploading to PyPi"
```
saartochner, line 44: upload the "previous" version to pypi, and not the one that we're writing now (https://lumigo.atlassian.net/browse/RD-3309)

src/test/unit/test_main_utils.py
```diff
@@ -234,60 +269,35 @@ def test_format_frame():
...
-    assert omit_keys(value) == output
+    assert variables["a"] == '"A"'
```
nirLumigo, line 272: The data here was changed, why? Do we support in the backend for this?

src/lumigo_tracer/utils.py
saartochner: yes.
This was just duplicated omitting...

src/lumigo_tracer/utils.py
```diff
@@ -426,3 +367,90 @@ def md5hash(d: dict) -> str:
...
+        d, is_truncated = omit_keys(d, max_size, regexes, enforce_jsonify)
+    elif isinstance(d, list):
+        size = 0
+        organs = []
+        for a in d:
+            organs.append(lumigo_dumps(a, max_size, regexes, enforce_jsonify))
+            size += len(organs[-1])
+            if size > max_size:
+                break
+        return "[" + ", ".join(organs) + "]"
```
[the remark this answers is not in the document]
saartochner: all the other types should be handled in the `json.dumps` one line below :)

src/lumigo_tracer/parsers/utils.py
saartochner: nope. 5 lines above we trim the tree if we get None. returning `{}` makes it to never be trimmed...

src/lumigo_tracer/spans_container.py
```diff
@@ -285,12 +284,14 @@ def create_span(cls, event=None, context=None, force=False) -> "SpansContainer":
         """
         if cls._span and not force:
             return cls._span
+        # copy the event to ensure that we will not change it
+        event = copy.deepcopy(event)
```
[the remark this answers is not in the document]
saartochner: yes, but it makes it correct.
It happens just once per invocation, and it strongly validates that we're not changing the event

src/test/unit/test_main_utils.py
saartochner: yes. This is actually more "jsony" correct. Now the BE will get this as a string

src/lumigo_tracer/utils.py
```diff
@@ -426,3 +366,118 @@ def md5hash(d: dict) -> str:
     h = hashlib.md5()
     h.update(json.dumps(d, sort_keys=True).encode())
     return h.hexdigest()
+
+
+class IntermediateOmitResult:
+    __slots__ = ["d", "free_space"]
```
nirLumigo, line 372: 🚀

src/lumigo_tracer/utils.py
```diff
@@ -369,62 +368,89 @@ def md5hash(d: dict) -> str:
     return h.hexdigest()
 
 
+class IntermediateOmitResult:
```
nirLumigo, line 371: It can be private, isn't it?
```suggestion
class _IntermediateOmitResult:
```

src/lumigo_tracer/utils.py
```diff
@@ -369,62 +368,89 @@ def md5hash(d: dict) -> str:
...
-        return d, size
+    if result.free_space <= 0:
+        return result
     if key in SKIP_SCRUBBING_KEYS:
-        d[key] = value
-        size += len(value) if isinstance(value, str) else len(json.dumps(value))
-    elif isinstance(key, str) and regexes and regexes.match(key):
-        d[key] = "****"
-        size += 4
+        result.d[key] = value
```
nirLumigo, line 399: This isn't immutable 🤔

src/lumigo_tracer/utils.py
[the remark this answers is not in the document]
nirLumigo: Shouldn't there be try-catch for the `json.dumps`, if `d` is `datetime` there will be an exception.

PR #127, opened 2020-07-26
RD-3341 - fix: missing uri in 'requests' package

saartochner:


src/lumigo_tracer/spans_container.py
```diff
@@ -150,13 +150,12 @@ def add_unparsed_request(self, parse_params: HttpRequest):
...
         if self.http_spans:
             last_event = self.http_spans[-1]
-            if last_event and last_event.get("type") == HTTP_TYPE:
+            if last_event and last_event.get("type") == HTTP_TYPE and self.previous_request:
                 if last_event.get("info", {}).get("httpInfo", {}).get("host") == parse_params.host:
                     if "response" not in last_event["info"]["httpInfo"]:
                         self.http_spans.pop()
-                        prev_headers, prev_body = self.previous_request
-                        body = (prev_body + parse_params.body)[:MAX_BODY_SIZE]
-                        self.add_request_event(parse_params.clone(headers=prev_headers, body=body))
```
saartochner, line 159: we didn't clone the `uri` from the previous request.
This is generally smells - we need to clone the previous request (that has all the data) and just override the new body.
That's what I did :)

src/lumigo_tracer/spans_container.py
```diff
@@ -150,13 +150,12 @@ def add_unparsed_request(self, parse_params: HttpRequest):
         """
         if self.http_spans:
             last_event = self.http_spans[-1]
-            if last_event and last_event.get("type") == HTTP_TYPE:
+            if last_event and last_event.get("type") == HTTP_TYPE and self.previous_request:
```
nirLumigo, line 153: What happens if this is the first request? We won't have `previous_request`, isn't it?

src/test/unit/sync_http/test_sync_hook.py
```diff
@@ -111,6 +111,21 @@ def lambda_test_function():
...
+    def lambda_test_function():
+        conn = http.client.HTTPConnection("www.google.com")
+        conn.request("POST", "/?q=123", b"123")
+        conn.send(BytesIO(b"456"))
+
+    lambda_test_function()
+    http_spans = SpansContainer.get_span().http_spans
+
+    assert http_spans
+    print(http_spans[0]["info"]["httpInfo"]["request"])
```
nirLumigo, line 125: Why print it?

src/test/unit/sync_http/test_sync_hook.py
```diff
@@ -111,6 +111,21 @@ def lambda_test_function():
     assert http_spans[0]["info"]["httpInfo"]["request"]["uri"] == "www.google.com/?q=123"
 
 
+def test_uri_requests():
+    @lumigo_tracer(token="123")
+    def lambda_test_function():
+        conn = http.client.HTTPConnection("www.google.com")
```
nirLumigo, line 117: Don't we use a package to mock the API call?

src/lumigo_tracer/spans_container.py
saartochner: excatly, and then we don't need to do this part.
Previously we handle it be giving default parameters that doesn't change anything - and I think that. this is better approach :)

src/test/unit/sync_http/test_sync_hook.py
saartochner: 😂 
Copied from other test and didn't look...
Fixed in the origin too :)

src/test/unit/sync_http/test_sync_hook.py
saartochner: nope.
like in the rest of the tests in this file, we're actually do API requests (just like our customers) and we check that the tracer parse them well :)

PR #129, opened 2020-07-31
reset connection to edge on error

saartochner:


src/lumigo_tracer/utils.py
```diff
@@ -199,12 +209,8 @@ def report_json(region: Union[None, str], msgs: List[dict]) -> int:
...
     duration = 0
     if not edge_connection or edge_connection.host != host:
-        try:
-            edge_connection = http.client.HTTPSConnection(  # type: ignore
-                host.lstrip("https://").rstrip(EDGE_PATH), timeout=EDGE_TIMEOUT
-            )
-        except Exception as e:
-            get_logger().exception(f"Could not establish connection to {host}", exc_info=e)
+        edge_connection = establish_connection(host)
+        if not edge_connection:
```
efimk-lu, line 213: It looks like you had a bug here before your change, returning duration although you had an error, am I right?

src/lumigo_tracer/utils.py
```diff
@@ -220,6 +226,7 @@ def report_json(region: Union[None, str], msgs: List[dict]) -> int:
             get_logger().info(f"successful reporting, code: {getattr(response, 'code', 'unknown')}")
         except Exception as e:
             get_logger().exception(f"Could not report json to {host}", exc_info=e)
+            edge_connection = establish_connection(host)
```
efimk-lu, line 229: Would have been nice to add to the log message (upper line), the word retrying.
It's not a full fix, right? this request is lost, the next one hopefully will succeed.

efimk-lu: From what I saw even "clean" lambdas fail, so not sure how recreating the connection will help

src/lumigo_tracer/utils.py
saartochner: nope. the function that calls it expects to get duration, and duration=0 is what returned in case of an error

src/lumigo_tracer/utils.py
saartochner: My guess is that it happens in warm lambdas, that loses their connection between the invocations somehow...
It is indeed not a full fix because I need to validate it, but that's absolutely a solid start

saartochner: fixed log message 👍

PR #138, opened 2020-08-10
allow decimal return value

saartochner:


src/lumigo_tracer/utils.py
```diff
@@ -157,7 +157,7 @@ def _is_span_has_error(span: dict) -> bool:
 
 
 def _get_event_base64_size(event) -> int:
-    return len(b64encode(json.dumps(event).encode()))
+    return len(b64encode(json.dumps(event, cls=DecimalEncoder).encode()))
```
nirLumigo, line 160: I think `json.dumps(event, cls=DecimalEncoder)` should be extracted to a function, because people might forget to add `DecimalEncoder` in the future. But we usually use a util if it exists.

saartochner: good point!

PR #139, opened 2020-08-17
RD-3434 accurate requests time

saartochner:
For the lambda:
image

Before:
image
After:
image

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,4 +5,4 @@ boto3
 pytest-cov==2.6.1
 capturer==2.4
 attrs==19.1.0
-requests
+requests==2.24.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,3 +5,4 @@ boto3
 pytest-cov==2.6.1
 capturer==2.4
 attrs==19.1.0
+requests
--- a/src/lumigo_tracer/sync_http/sync_hook.py
+++ b/src/lumigo_tracer/sync_http/sync_hook.py
@@ -1,3 +1,4 @@
+from datetime import datetime
 import inspect
 import logging
 import http.client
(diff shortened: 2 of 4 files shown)


requirements.txt
```diff
@@ -5,3 +5,4 @@ boto3
 pytest-cov==2.6.1
 capturer==2.4
 attrs==19.1.0
+requests
```
nirLumigo, line 8: Why not pin the version?

src/lumigo_tracer/sync_http/sync_hook.py
```diff
@@ -313,4 +328,6 @@ def wrap_http_calls():
                 wrap_function_wrapper(
                     "urllib3.response", "HTTPResponse.read_chunked", _read_stream_wrapper
                 )
+            if importlib.util.find_spec("requests"):
```
nirLumigo, line 331: 🚀

src/lumigo_tracer/sync_http/sync_hook.py
```diff
@@ -313,4 +328,6 @@ def wrap_http_calls():
...
+                wrap_function_wrapper("requests.api", "request", _requests_wrapper)
```
nirLumigo, line 332: Won't there a double wrapping for the request?

src/test/unit/sync_http/test_sync_hook.py
```diff
@@ -502,6 +503,27 @@ def lambda_test_function(event, context):
...
+        start_time = time.time() * 1000
+        requests.get("https://www.google.com")
+        return start_time
+
+    # add delay to the connection establishment process
+    original_getaddrinfo = socket.getaddrinfo
+    monkeypatch.setattr(
+        socket,
+        "getaddrinfo",
+        lambda *args, **kwargs: time.sleep(0.1) or original_getaddrinfo(*args, **kwargs),
```
nirLumigo, line 518: I think it should be extracted to a function, I don't think we should combine lines by putting `or` between them, makes it hard to read.

src/lumigo_tracer/sync_http/sync_hook.py
saartochner: there is!
Note that `_requests_wrapper` (the "higher" hook) doesn't create span by itself, but only append data to the span that was created with `http.client` (the lower hook).

PR #140, opened 2020-08-17
RD-3449 - fix missing URI when avoid parsing headers

saartochner:
Before:
image

After:
image


src/lumigo_tracer/sync_http/sync_hook.py
```diff
@@ -31,6 +32,9 @@
 LUMIGO_HEADERS_HOOK_KEY = "_lumigo_headers_hook"
 
 
+HookedData = namedtuple("HookedData", ["headers", "path"])
```
nirLumigo, line 35: Why not `dataclass`?

saartochner: better performance :(

PR #146, opened 2020-09-21
RD-3661 - send start span on timeout if send_only_on_error

saartochner:


src/lumigo_tracer/spans_container.py
```diff
@@ -116,6 +121,8 @@ def start(self, event=None, context=None):
     def handle_timeout(self, *args):
         get_logger().info("The tracer reached the end of the timeout timer")
         to_send = [s for s in self.http_spans if s["id"] in self.http_span_ids_to_send]
+        if Configuration.send_only_if_error:
+            to_send.append(self._generate_start_span())
```
nirLumigo, line 125: Does the order matter? Maybe it should be the first in `to_send`?
`to_send.insert(0, self._generate_start_span())`

src/test/unit/test_spans_container.py
```diff
@@ -83,6 +78,23 @@ def test_spans_container_end_function_send_only_on_errors_mode_false_not_effecti
...
+    SpansContainer.create_span()
+    SpansContainer.get_span().start()
+    SpansContainer.get_span().add_request_event(dummy_http_request)
+
+    SpansContainer.get_span().handle_timeout()
+
+    messages = reporter_mock.call_args_list[0][1]["msgs"]
+    assert len(messages) == 2
+    assert [m for m in messages if m["type"] == FUNCTION_TYPE and m["id"].endswith("_started")]
+    assert [m for m in messages if m["type"] == HTTP_TYPE]
```
nirLumigo, line 95: Shouldn't there be also an `end_span`?

src/test/unit/test_spans_container.py
```diff
@@ -83,6 +78,23 @@ def test_spans_container_end_function_send_only_on_errors_mode_false_not_effecti
...
+):
+    monkeypatch.setattr(Configuration, "send_only_if_error", True)
+
+    SpansContainer.create_span()
+    SpansContainer.get_span().start()
+    SpansContainer.get_span().add_request_event(dummy_http_request)
+
+    SpansContainer.get_span().handle_timeout()
+
+    messages = reporter_mock.call_args_list[0][1]["msgs"]
```
nirLumigo, line 92: If you can find a better way to extract the messages, it will be nice :)
Maybe you can use the pattern we have in testing-utils. Not a must.

nirLumigo (changes requested):
Nice :)

src/lumigo_tracer/spans_container.py
saartochner: it doesn't metter, and if there will be many spans it is unnecessary compute waste 🤷

src/test/unit/test_spans_container.py
saartochner: nope!
This is the TO mechanism - we give the lambda a chance to finish by itself.

nirLumigo (approved):

PR #147, opened 2020-09-24
adding s3 & cloudfront event parsers

GuyMoses:


src/test/unit/parsers/test_event_parser.py
```diff
@@ -404,3 +411,165 @@ def test_parse_event_sqs():
...
+                "Records": [
+                    {
+                        "awsRegion": "us-west-2",
+                        "eventTime": "2020-09-24T12:00:12.137Z",
+                        "eventName": "ObjectCreated:Put",
+                        "userIdentity": {"principalId": "A2QVTU9T5VMOU3"},
+                        "requestParameters": {"sourceIPAddress": "77.127.93.97"},
+                        "s3": {
+                            "bucket": {
+                                "name": "testingbuckets3testing",
```
uri-p, line 434: Seems like this is redundant with arn

uri-p: Maybe we can also drop ownerIdentity

src/test/unit/parsers/test_event_parser.py
```diff
@@ -404,3 +411,165 @@ def test_parse_event_sqs():
...
+                                "ownerIdentity": {"principalId": "A2QVTU9T5VMOU3"},
+                                "arn": "arn:aws:s3:::testingbuckets3testing",
+                            },
+                            "object": {
+                                "key": "Screen+Shot+2020-05-27+at+12.37.36.png",
+                                "size": 61211,
+                                "eTag": "714ee5196a5c0a6e6b9019caa7b6e970",
```
uri-p, line 441: What is etag ? sequencer?
Can we drop them?

uri-p (changes requested):

uri-p (approved):

doriaviram (approved):

PR #152, opened 2020-10-11
Rd 3761 wrap pymongo

saartochner:


requirements.txt
```diff
@@ -6,3 +6,4 @@ pytest-cov==2.6.1
 capturer==2.4
 attrs==19.1.0
 requests==2.24.0
+pymongo==3.11.0
```
doriaviram, line 9: This is not bundled into the python tracer I guess, right?

src/lumigo_tracer/wrappers/pymongo/pymongo_wrapper.py
```diff
@@ -0,0 +1,63 @@
...
+                last_span = SpansContainer.get_span().get_last_span()
+                if not last_span or last_span.get("mongoRequestId") != event.request_id:
+                    get_logger().warning("Mongo span ended without a record on its start")
+                    return
+                last_span.update(
+                    {
+                        "ended": last_span["started"] + (event.duration_micros / 1000),
+                        "response": lumigo_dumps(event.reply),
+                    }
+                )
```
doriaviram, line 43: We got the spaned anyway, why just not get it by the ID? it's way better than, last span way

src/test/unit/wrappers/pymongo/test_pymongo_wrapper.py
```diff
@@ -0,0 +1,64 @@
...
+        connection_id="cid",
+    )
+
+
+@pytest.fixture
+def success_event():
+    return SimpleNamespace(duration_micros=5000, reply={"code": 200}, request_id="rid")
+
+
+@pytest.fixture
```
doriaviram, line 26: Missing tests:
- pymongo not exist

requirements.txt
```diff
@@ -6,3 +6,4 @@ pytest-cov==2.6.1
 capturer==2.4
```
doriaviram, line 6: General notes before merging:
- Check this on auto-instrument (in node we had issues over there)

doriaviram (changes requested):

requirements.txt
saartochner: right :)
This is just dev dependencies

requirements.txt
saartochner: will do! Thanks :)

doriaviram (approved):

PR #155, opened 2020-10-14
Double envs and event size limit on error

nirLumigo:
Example:
The event is longer than 3000 chars
On lambda success:
image
On lambda exception:
image

### DIFF
--- a/src/lumigo_tracer/lumigo_utils.py
+++ b/src/lumigo_tracer/lumigo_utils.py
@@ -55,6 +55,7 @@ NUMBER_OF_SPANS_IN_REPORT_OPTIMIZATION = 200
 DEFAULT_KEY_DEPTH = 4
 LUMIGO_TOKEN_KEY = "LUMIGO_TRACER_TOKEN"
 KILL_SWITCH = "LUMIGO_SWITCH_OFF"
+ERROR_SIZE_LIMIT_MULTIPLIER = 2
 
 _logger: Dict[str, logging.Logger] = {}
 
@@ -87,7 +88,7 @@ class Configuration:
     @staticmethod
     def get_max_entry_size(has_error: bool = False) -> int:
         if has_error:
-            return Configuration.max_entry_size * 2
+            return Configuration.max_entry_size * ERROR_SIZE_LIMIT_MULTIPLIER
         return Configuration.max_entry_size
 
 
--- a/src/lumigo_tracer/lumigo_utils.py
+++ b/src/lumigo_tracer/lumigo_utils.py
@@ -84,6 +84,12 @@ class Configuration:
     max_entry_size: int = DEFAULT_MAX_ENTRY_SIZE
     get_key_depth: int = DEFAULT_KEY_DEPTH
 
+    @staticmethod
+    def get_max_entry_size(has_error: bool = False) -> int:
+        if has_error:
+            return Configuration.max_entry_size * 2
+        return Configuration.max_entry_size
+
 
 def config(
     edge_host: str = "",
--- a/src/lumigo_tracer/spans_container.py
+++ b/src/lumigo_tracer/spans_container.py
@@ -205,7 +205,7 @@ class SpansContainer:
     def add_tag(self, key: str, value: str) -> None:
         self.function_span[EXECUTION_TAGS_KEY].append({"key": key, "value": value})
 
-    def end(self, ret_val=None) -> Optional[int]:
+    def end(self, event: Optional[dict], ret_val=None) -> Optional[int]:
         TimeoutMechanism.stop()
         reported_rtt = None
         self.previous_request = None
@@ -226,6 +226,12 @@ class SpansContainer:
                     "The lambda will probably fail due to bad return value. " + suffix,
                 )
         self.function_span.update({"return_value": parsed_ret_val})
+        if _is_span_has_error(self.function_span):
+            self.function_span["envs"] = _get_envs_for_span(has_error=True)
+            if event:
+                self.function_span["event"] = EventDumper.dump_event(
+                    copy.deepcopy(event), has_error=True
+                )
         spans_contain_errors: bool = any(
             _is_span_has_error(s) for s in self.spans + [self.function_span]
         )
@@ -266,7 +272,7 @@ class SpansContainer:
         additional_info = {}
         if Configuration.verbose:
             additional_info.update(
-                {"event": EventDumper.dump_event(event), "envs": lumigo_dumps(dict(os.environ))}
+                {"event": EventDumper.dump_event(event), "envs": _get_envs_for_span()}
             )
 
         trace_root, transaction_id, suffix = parse_trace_id(os.environ.get("_X_AMZN_TRACE_ID", ""))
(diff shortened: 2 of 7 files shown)


efimk-lu:
@saartochner for some unknown reason all the IT pass to @nirLumigo on the first push ;-)

src/lumigo_tracer/spans_container.py
```diff
@@ -226,6 +226,12 @@ def end(self, ret_val=None) -> Optional[int]:
                     "The lambda will probably fail due to bad return value. " + suffix,
                 )
         self.function_span.update({"return_value": parsed_ret_val})
+        if _is_span_has_error(self.function_span):
+            self.function_span["envs"] = _get_envs_for_span(has_error=True)
+            if event:
+                self.function_span["event"] = EventDumper.dump_event(
+                    copy.deepcopy(event), has_error=True
```
nirLumigo, line 233: I'm doing here `deepcopy` because this is what we do in the function `create_span`.
The other option is to `deepcopy` at the calling functions, I think it will be worse because this means we'll keep a copy of the event in the memory for the entire execution of lambda although in most cases the event will be used only at the beginning.

src/lumigo_tracer/lumigo_utils.py
```diff
@@ -84,6 +84,12 @@ class Configuration:
     max_entry_size: int = DEFAULT_MAX_ENTRY_SIZE
     get_key_depth: int = DEFAULT_KEY_DEPTH
 
+    @staticmethod
+    def get_max_entry_size(has_error: bool = False) -> int:
+        if has_error:
+            return Configuration.max_entry_size * 2
```
saartochner, line 90: don't you prefer that the `2` will be configurable as well?
At least let it be a constant :-)

src/lumigo_tracer/tracer.py
```diff
@@ -66,7 +66,8 @@ def lambda_wrapper(*args, **kwargs):
                     SpansContainer.get_span().add_exception_event(e, inspect.trace())
                 raise
             finally:
-                SpansContainer.get_span().end(ret_val)
+                event = args[0] if len(args) > 0 else None
```
saartochner, line 69: this logic should not be here.
`SpansContainer` should remember the event by itself

src/lumigo_tracer/spans_container.py
saartochner: agree that it should be here.

I'm not sure about its relevance - because the function has already been finished, so what's the point of not changing the event?
On the other hand, some other "wrapper" could run after us, and we might change this value 🤔

src/test/unit/test_spans_container.py
```diff
@@ -82,6 +83,22 @@ def test_spans_container_end_function_send_only_on_errors_mode_false_not_effecti
...
+    event = {"k": long_string}
+    SpansContainer.create_span(event)
+    SpansContainer.get_span().start()
+    start_span = copy.deepcopy(SpansContainer.get_span().function_span)
+    SpansContainer.get_span().add_exception_event(Exception("Some Error"), inspect.trace())
+
+    SpansContainer.get_span().end(event)
+
+    end_span = SpansContainer.get_span().function_span
+    assert len(end_span["event"]) > len(start_span["event"]) * 1.8 + 1
```
saartochner, line 98: why do you compare to the start span?
You built this string in the test itself - use `long_string` explicitly,
and then you can drop this weird formula `*1.8+1`

src/test/unit/event/test_event_dumper.py
```diff
@@ -479,6 +479,13 @@ def test_parse_cloudfront_event(cloudfront_event):
     )
 
 
+def test_dump_event_has_error_should_double_limit_size():
+    event = {"k": "v" * Configuration.get_max_entry_size() * 3}
+    result_no_error = EventDumper.dump_event(event)
+    result_has_error = EventDumper.dump_event(event, has_error=True)
+    assert len(result_has_error) > len(result_no_error) * 1.8 + 1
```
saartochner, line 486: again, use `event["k"]` explicitly - check exactly what you want to see

src/lumigo_tracer/spans_container.py
```diff
@@ -266,7 +272,7 @@ def create_span(cls, event=None, context=None, force=False) -> "SpansContainer":
         additional_info = {}
         if Configuration.verbose:
             additional_info.update(
-                {"event": EventDumper.dump_event(event), "envs": lumigo_dumps(dict(os.environ))}
```
saartochner, line 269: I think that it better to just pass `event` to SpansContainer, and then the `envs` and `event` could be calculated inside the instance 🤔

saartochner (changes requested):
COOL!

src/lumigo_tracer/spans_container.py
saartochner: don't you want to do this logic even if one of the HTTPs has failed?
take it one line below and use `spans_contain_errors`.
Moreover, this function gets fat. Can you extract it to a method?

src/lumigo_tracer/spans_container.py
```diff
@@ -226,6 +226,12 @@ def end(self, ret_val=None) -> Optional[int]:
...
         self.function_span.update({"return_value": parsed_ret_val})
+        if _is_span_has_error(self.function_span):
+            self.function_span["envs"] = _get_envs_for_span(has_error=True)
+            if event:
+                self.function_span["event"] = EventDumper.dump_event(
+                    copy.deepcopy(event), has_error=True
+                )
         spans_contain_errors: bool = any(
             _is_span_has_error(s) for s in self.spans + [self.function_span]
         )
```
saartochner, line 237: I think that `lumigo_dumps` will trim your string again.
please check it E2E. I don't think that it works...

src/lumigo_tracer/spans_container.py
nirLumigo: About the relevance of the event - there are other cases as well, like returning the event as `return_value` of the lambda.

About "don't you want to do this logic even if one of the HTTPs has failed?" - I consulted with Efi and we decided to do it only if the lambda fails.

src/lumigo_tracer/spans_container.py
nirLumigo: It works :)
image
image

Where is the `lumigo_dumps` that might trim my string?

src/lumigo_tracer/spans_container.py
saartochner: ohh of course that only if lambda fails,
But why don't you just pass it to the SpansContainer - and it will call the EventDumper by itself?

src/lumigo_tracer/spans_container.py
nirLumigo: I don't understand what do you mean. Do you mean that it should be a function inside the `SpansContainer` that updates `event` and `envs`?
In this part of the code, `SpansContainer` still doesn't exist.

saartochner: I know, it will be initialized a few lines later.
I meant to pass the event to the SpansContainer as a parameter, and then you can avoid the parsing in `tracer.py`

src/test/unit/test_spans_container.py
nirLumigo: I need to compare to the start span because I want to make sure that we send extra data in case of an error.
For example, someone will change the size_limit to 5000 and in case of an error we won't double the size_limit. In this case I want the test to fail

saartochner (approved):

PR #158, opened 2020-10-20
RD-3862 - feat: support eventBridge

saartochner:


src/lumigo_tracer/event/event_trigger.py
```diff
@@ -118,6 +120,23 @@ def _parse_sns(event: dict):
...
+def _is_event_bridge(event: dict):
+    return (
+        "version" in event
+        and "id" in event  # noqa: W503
+        and "detail-type" in event  # noqa: W503
+        and "source" in event  # noqa: W503
+        and "time" in event  # noqa: W503
+        and "region" in event  # noqa: W503
+        and "resources" in event  # noqa: W503
+        and "detail" in event  # noqa: W503
```
doriaviram, line 132: Looks too generic, try to find something deeper that related to the event bridge

src/lumigo_tracer/wrappers/http/http_parser.py
```diff
@@ -234,6 +234,38 @@ def parse_response(self, url: str, status_code: int, headers, body: bytes) -> di
...
+        except json.JSONDecodeError as e:
+            get_logger().debug("Error while trying to parse eventBridge request body", exc_info=e)
+            parsed_body = {}
+        resource_names = set()
+        if isinstance(parsed_body.get("Entries"), list):
+            resource_names = {
+                e["EventBusName"] for e in parsed_body["Entries"] if e.get("EventBusName")
+            }
+        return recursive_json_join(
+            {"info": {"resourceName": ",".join(resource_names) or None}},
```
doriaviram, line 250: This is our way to pass lists to the backend? cant we use array?

doriaviram (changes requested):

src/lumigo_tracer/event/event_trigger.py
saartochner: Unfortunately from the event size we have nothing..
Maybe I can check also the types of these fields, but the was infro doesn't give me anything...
This is how it looks:
image

src/lumigo_tracer/wrappers/http/http_parser.py
saartochner: So that's actually a string - I don't want the BE to handle it as a list 🤔

src/lumigo_tracer/event/event_trigger.py
saartochner: wdyt? @doriaviram

src/lumigo_tracer/wrappers/http/http_parser.py
doriaviram: Its a list of all the event buses, you just make it a string by joining with `,`
Better option: pass another field `resourceNames` to the backend, the make this join over there when building the span.
The main issue here, its if we will want to show that in another way to our users (as a list for example), we will not able to do-it because this code in our tracer can't be really change (we will need all ways backward compatible)

saartochner: Awesome idea!
Will do!

doriaviram (approved):

PR #161, opened 2020-10-27
fix: http uri parsing

GuyMoses:


src/lumigo_tracer/lumigo_utils.py
```diff
@@ -544,3 +544,11 @@ def get_size_upper_bound() -> int:
 
 def is_error_code(status_code: int) -> bool:
     return status_code >= 400
+
+
+def is_arn(string_to_validate: str) -> bool:
+    return string_to_validate.startswith("arn:aws:")
+
+
+def extract_name_from_arn(arn: str) -> str:
```
saartochner, line 553: the 6'th index is not always the name - you assume something on the arn. 
Is that a function arn? if so, please write it in the function's name

src/lumigo_tracer/lumigo_utils.py
```diff
@@ -544,3 +544,11 @@ def get_size_upper_bound() -> int:
 
 def is_error_code(status_code: int) -> bool:
     return status_code >= 400
+
+
+def is_arn(string_to_validate: str) -> bool:
```
saartochner, line 549:
```suggestion
def is_aws_arn(string_to_validate: str) -> bool:
```

src/lumigo_tracer/wrappers/http/http_parser.py
```diff
@@ -170,9 +173,14 @@ def parse_response(self, url: str, status_code: int, headers, body: bytes) -> di
 
 class LambdaParser(ServerlessAWSParser):
     def parse_request(self, parse_params: HttpRequest) -> dict:
+        decoded_uri = safe_split_get(unquote(parse_params.uri), "/", 3)
```
saartochner, line 176: Just checked- it can return None (can you update the signature?)
 make sure to handle it in `is_arn`

src/test/unit/test_lumigo_utils.py
```diff
@@ -409,3 +411,16 @@ def test_get_size_upper_bound():
 )
 def test_is_error_code(status_code, is_error):
     assert is_error_code(status_code) is is_error
+
+
+@pytest.mark.parametrize(
+    ("arn", "is_arn_result"),
+    [("not-arn", False), ("arn:aws:lambda:region:876841109798:function:function-name", True,)],
```
saartochner, line 418: add test for non str inputs :)

saartochner (changes requested):
Well done!

src/lumigo_tracer/wrappers/http/http_parser.py
GuyMoses: weird - it returns str according to PyCharm

saartochner: you're right. the function's signature is bad. Can you fix it please? :)

src/lumigo_tracer/lumigo_utils.py
```diff
@@ -544,3 +544,11 @@ def get_size_upper_bound() -> int:
...
 def is_error_code(status_code: int) -> bool:
     return status_code >= 400
+
+
+def is_aws_arn(string_to_validate: Optional[str]) -> bool:
+    return bool(string_to_validate and string_to_validate.startswith("arn:aws:"))
+
+
+def extract_function_name_from_arn(arn: str) -> str:
+    return arn.split(":")[6]
```
saartochner, line 554: you can use `safe_split_get` (find an example in this file), which is safer

saartochner (approved):
Awesome! 
Approve in advanced :)

PR #162, opened 2020-11-08
adding appsync trigger-by

GuyMoses:


src/lumigo_tracer/event/event_trigger.py
```diff
@@ -133,10 +135,22 @@ def _is_event_bridge(event: dict):
     )
 
 
+def _is_appsync(event: dict) -> bool:
+    host = event.get("context", {}).get("request", {}).get("headers", {}).get("host")
```
saartochner, line 139: `safe_get(event, ["context", "request", "headers", "host"])`
(your code will break on `event = {"context": "a"}`)

src/lumigo_tracer/event/event_trigger.py
```diff
@@ -133,10 +135,22 @@ def _is_event_bridge(event: dict):
...
+    return host and "appsync-api" in host
```
saartochner, line 140: `and isinstance(host, str)`
(your code will break on `host=5`)

src/lumigo_tracer/event/event_trigger.py
```diff
@@ -133,10 +135,22 @@ def _is_event_bridge(event: dict):
...
+
+
 def _parse_event_bridge(event: dict):
     return {"triggeredBy": "eventBridge", "messageId": event["id"]}
 
 
+def _parse_appsync(event: dict) -> dict:
+    headers = event.get("context", {}).get("request", {}).get("headers", {})
```
saartochner, line 148: same as before

src/lumigo_tracer/event/event_trigger.py
saartochner: `and '.' in host`

src/lumigo_tracer/event/event_trigger.py
```diff
@@ -133,10 +135,22 @@ def _is_event_bridge(event: dict):
...
+    api_id = headers.get("host", "").split(".")[0]
```
saartochner, line 149: @doriaviram wdyt? Should we prefer to do this parsing in the BE?

src/lumigo_tracer/event/event_trigger.py
```diff
@@ -133,10 +135,25 @@ def _is_event_bridge(event: dict):
...
+
+
 def _parse_event_bridge(event: dict):
     return {"triggeredBy": "eventBridge", "messageId": event["id"]}
 
 
+def _parse_appsync(event: dict) -> dict:
+    headers = safe_get(event, ["context", "request", "headers"])
+    host = headers.get("host")
+    api_id = host.split(".")[0] if host and isinstance(host, str) else None
```
saartochner, line 150:
```suggestion
    api_id = host.split(".")[0]
```
You can assume it :)

GuyMoses: can I assume it's a string?

saartochner: I think - you checked it in `_is_appsync`...

GuyMoses: oh yeah, but code pieces could change for some reason

saartochner: 👍

saartochner (approved):
Well done!
Let's just wait for Dori's suggestion on the tracer-BE separation.
The code is approved on my side

PR #164, opened 2020-12-08
Support China region

nirLumigo:


src/lumigo_tracer/logger.py
```diff
@@ -0,0 +1,29 @@
+import logging
```
nirLumigo, line 1: All the code in this file was moved from lumigo_utils.py

src/lumigo_tracer/lumigo_utils.py
```diff
@@ -246,48 +279,45 @@ def report_json(region: Union[None, str], msgs: List[dict], should_retry: bool =
...
+        if should_retry:
+            get_logger().exception(f"Could not report to {host}. Retrying.", exc_info=e)
+            edge_connection = establish_connection(host)
+            report_json(region, msgs, should_retry=False)
+        else:
+            get_logger().exception("Could not report: A span was lost.", exc_info=e)
     return duration
 
 
-def get_logger(logger_name="lumigo"):
```
nirLumigo, line 271: This function was moved to a new file: logger.py

src/lumigo_tracer/kinesis_service.py
```diff
@@ -0,0 +1,163 @@
+import random
```
saartochner, line 1: move to libs

src/lumigo_tracer/boto.py
```diff
@@ -0,0 +1,31 @@
+from typing import Optional
+
+from lumigo_tracer.logger import get_logger
+
+try:
+    import botocore
+except Exception:
+    get_logger().warning("botocore is missing")
```
saartochner, line 8: This is ok that there is no boto

src/lumigo_tracer/kinesis_service.py
```diff
@@ -0,0 +1,163 @@
...
+import json
+from collections import Counter
+from typing import Callable, Any, List, Dict, Tuple, Optional
+
+from lumigo_tracer import boto
```
saartochner, line 6: he shouldn't know the tracer. this is a util

saartochner (changes requested):
GJ!

src/test/unit/test_tracer.py
```diff
@@ -364,3 +370,57 @@ def lambda_test_function(event, context):
     # following python's runtime: runtime/lambda_runtime_marshaller.py:27
     expected_message = 'The lambda will probably fail due to bad return value. Original message: "Object of type datetime is not JSON serializable"'
     assert function_span["error"]["message"] == expected_message
+
+
+@mock_kinesis
+def test_china(context, reporter_mock, monkeypatch):
+    china_region_for_test = "ap-east-1"  # Moto doesn't work for China
+    monkeypatch.setattr(lumigo_utils, "CHINA_REGION", china_region_for_test)
+    monkeypatch.setenv("AWS_REGION", china_region_for_test)
```
saartochner, line 379: IMO you don't really need these lines, as long as line 385 will use the fake region 🤔

saartochner (changes requested):
GJ!
It starts to look very good :)

src/test/unit/test_tracer.py
```diff
@@ -364,3 +370,57 @@ def lambda_test_function(event, context):
...
+    result = lambda_test_function(event, context)
+
+    assert result == "ret_value"
+    # Spans sent to Kinesis
+    records = client.get_records(ShardIterator=shard_iterator)["Records"]
+    assert len(records) == 2  # Start span and end span
+    span_sent = json.loads(records[1]["Data"].decode())[0]
+    assert span_sent["event"] == json.dumps(event)
+    # Used the client from the decorator params
+    kinesis_service.get_boto_client.assert_called_with(
```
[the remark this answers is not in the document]
nirLumigo: Moto doesn't check which user did the call and it doesn't check the credentials. I want to make sure that it uses the credentials passed to the tracer decorator

src/test/unit/test_tracer.py
nirLumigo: We need this because in `report_json` there is:
```
    if region == CHINA_REGION:
        return _publish_spans_to_kinesis(to_send, CHINA_REGION)
```
And the env variable will set the `region` variable.

When I tried to remove it the test failed because the tracer tried to send the spans via http (the old way) instead of pushing them to Kinesis

src/lumigo_tracer/lumigo_utils.py
```diff
@@ -235,9 +260,19 @@ def report_json(region: Union[None, str], msgs: List[dict], should_retry: bool =
...
     """
-    global edge_connection
+    if not Configuration.should_report:
+        return 0
     get_logger().info(f"reporting the messages: {msgs[:10]}")
+    try:
+        prune_trace: bool = not os.environ.get("LUMIGO_PRUNE_TRACE_OFF", "").lower() == "true"
+        to_send = _create_request_body(msgs, prune_trace).encode()
+    except Exception as e:
+        get_logger().exception("Failed to create request: A span was lost.", exc_info=e)
```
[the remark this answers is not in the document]
nirLumigo: Are you sure about it? We do it in different places in the code:
image

image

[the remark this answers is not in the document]
nirLumigo: Looks like both options are ok: (https://docs.python.org/3.7/library/logging.html)
"If exc_info does not evaluate as false, it causes exception information to be added to the logging message. If an exception tuple (in the format returned by sys.exc_info()) or an exception instance is provided, it is used; otherwise, sys.exc_info() is called to get the exception information."

src/test/unit/test_tracer.py
saartochner: yeah, but what if you'll just use the real china region?
the only reason not to do it is moto, 
and you're mocking it either way 🤷

src/lumigo_tracer/lumigo_utils.py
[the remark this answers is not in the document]
saartochner: awesome, good to learn! :)

saartochner (approved):
awesome!
Just check if my comment on the test makes sense,
otherwise, it's approved :)

PR #166, opened 2020-12-14
RD-4202 - prevent collection of lumigo events

saartochner:


src/lumigo_tracer/spans_container.py
```diff
@@ -115,10 +115,10 @@ def start(self, event=None, context=None):
     def handle_timeout(self, *args):
         get_logger().info("The tracer reached the end of the timeout timer")
         to_send = [s for s in self.spans if s["id"] in self.span_ids_to_send]
+        self.span_ids_to_send.clear()
```
saartochner, line 118: this fixes the first bug: the last HTTP call looks infinite although it has finished.
It happens because this call finishes while we send the TO report. I.e. we clean the `self.span_ids_to_send` too late.

src/lumigo_tracer/spans_container.py
```diff
@@ -158,7 +158,9 @@ def update_event_end_time(self) -> None:
         This function assumes synchronous execution - we update the last http event.
         """
         if self.spans:
-            self.spans[-1]["ended"] = get_current_ms_time()
+            span = self.spans[-1]
+            span["ended"] = get_current_ms_time()
+            self.span_ids_to_send.add(span["id"])
```
saartochner, line 163: If we cleaned the `self.span_ids_to_send` - we should resend this span

src/lumigo_tracer/wrappers/http/sync_http_wrappers.py
```diff
@@ -29,10 +30,19 @@
 HookedData = namedtuple("HookedData", ["headers", "path"])
 
 
+def is_lumigo_edge(host: Optional[str]):
```
saartochner, line 33: This solves the second bug: when we catch our own spans.
We just need to ignore them.

src/lumigo_tracer/lumigo_utils.py
```diff
@@ -246,15 +246,16 @@ def establish_connection(host):
     return None
 
 
-def prepare_host(host):
+def get_edge_host(region: Optional[str] = None):
```
nirLumigo, line 249: Add return type

src/lumigo_tracer/lumigo_utils.py
```diff
@@ -246,15 +246,16 @@ def establish_connection(host):
...
+    host = Configuration.host or EDGE_HOST.format(region=region or os.environ.get("AWS_REGION"))
```
nirLumigo, line 250: I think that we already do `os.environ.get("AWS_REGION")` somewhere, maybe extract it to a function

src/lumigo_tracer/wrappers/http/sync_http_wrappers.py
nirLumigo, line 33:
```suggestion
def is_lumigo_edge(host: Optional[str]) -> bool:
```

nirLumigo (approved):

PR #167, opened 2020-12-15
RD-4242 - handle malformed trace id

saartochner:


src/lumigo_tracer/spans_container.py
```diff
@@ -260,8 +266,13 @@ def _set_error_extra_data(self, event):
             )
 
     def get_patched_root(self):
+        """
+        We're changing the root in order to find the component of the transaction faster. More info:
```
nirLumigo, line 270:
```suggestion
        We're changing the root in order to find the pass the transaction_id. More info:
```
Isn't it? 🤔

nirLumigo (approved):

PR #172, opened 2020-12-21
wrap requests only if kill switch is off

saartochner:


src/lumigo_tracer/spans_container.py
```diff
@@ -264,6 +264,9 @@ def _set_error_extra_data(self, event):
                 copy.deepcopy(event), has_error=True
             )
 
+    def can_path_root(self):
```
nirLumigo, line 267:
```suggestion
    def can_path_root(self) -> bool:
```

nirLumigo (approved):

PR #173, opened 2020-12-24
Rd 4241 unscrub aws keys

nirLumigo:
1. Adding scrubbing for list of dicts.
2. `lumigo_dumps` can skip scrubbing for specific paths.
3. Shouldn't scrub known AWS DynamoDB Key (getItem), like the body in the image:
image
4. Shouldn't scrub known AWS S3 Key (triggerBy), like the event in the image:
image
-------------
After the fix:
3.
image

4.
image


src/test/unit/event/test_event_dumper.py
```diff
@@ -418,7 +418,7 @@ def test_is_s3_event(s3_event):
 
 def test_parse_s3_event(s3_event):
     ordered_s3_event = EventDumper.dump_event(event=s3_event)
-    assert ordered_s3_event == lumigo_dumps(
+    assert ordered_s3_event == json.dumps(
```
nirLumigo, line 421: This tests that no scrubbing occurred because json.dumps doesn't do `omit_keys`

src/lumigo_tracer/wrappers/http/http_data_classes.py
```diff
@@ -8,13 +8,15 @@ class HttpRequest:
     uri: str
     headers: dict
     body: bytes
+    omit_skip_path: Optional[List[str]]
```
nirLumigo, line 11: We need to keep the path also here because when there is an error response code from an http request, we double the payload size and when we do it we also do `lumigo_dumps`

src/lumigo_tracer/event/event_dumper.py
```diff
@@ -221,7 +235,13 @@ def dump_event(
         for handler in handlers:
             try:
                 if handler.is_supported(event):
-                    return lumigo_dumps(handler.parse(event), max_size)
+                    return lumigo_dumps(
+                        handler.parse(event),
+                        max_size,
+                        omit_skip_path=None
+                        if should_scrub_known_services()
+                        else handler.get_omit_skip_path(),
```
saartochner, line 243: probably the `if` should be inside the `lumigo_dumps` function.
It seems like anyone else that will use this function will rewrite this if, right?

src/lumigo_tracer/lumigo_utils.py
```diff
@@ -66,6 +66,7 @@
 ERROR_SIZE_LIMIT_MULTIPLIER = 2
 CHINA_REGION = "cn-northwest-1"
 EDGE_KINESIS_STREAM_NAME = "prod_trc-inges-edge_edge-kinesis-stream"
+Container = TypeVar("Container", dict, list)
```
saartochner, line 69:
```suggestion
Container = NewType("Container", Union[List, Dict])
```
IMO :)
https://stackoverflow.com/questions/58755948/what-is-the-difference-between-typevar-and-newtype

src/lumigo_tracer/lumigo_utils.py
```diff
@@ -547,21 +549,38 @@ def _recursive_omitting(
...
         d[key] = float(value)
         free_space -= 5
+    elif isinstance(value, list):
+        d[key], free_space = reduce(
+            lambda p, i: _recursive_omitting(
+                p, (-1, i), regex, enforce_jsonify, omit_skip_path=omit_skip_path
+            ),
+            value,
+            ([], free_space),
+        )
```
saartochner, line 583: I don't really understand what you did here. it looks like you tried to build a new logic on the existing function, although it just makes it more complex (relevant to also lines 553-555).
What's the idea of appending None at the end? what is the meaning of key=-1?
Here's my idea:
This function should never get a list as it's input. 
Instead, you need to create another function, "list_recursive_omitting", that will handle lists. Then, these two functions should call one another when it's relevant.
The code will be much more clean and easy to understand IMO

src/lumigo_tracer/lumigo_utils.py
```diff
@@ -661,3 +689,7 @@ def is_error_code(status_code: int) -> bool:
 
 def is_aws_arn(string_to_validate: Optional[str]) -> bool:
     return bool(string_to_validate and string_to_validate.startswith("arn:aws:"))
+
+
+def should_scrub_known_services() -> bool:
```
saartochner, line 694: Can you use the `Configuration` class at the beginning?
this is our new standard :)

src/lumigo_tracer/wrappers/http/http_parser.py
```diff
@@ -43,9 +44,14 @@ class Parser:
 
     def parse_request(self, parse_params: HttpRequest) -> dict:
         if Configuration.verbose and parse_params and not should_scrub_domain(parse_params.host):
+            parse_params.omit_skip_path = (
+                None if should_scrub_known_services() else self.get_omit_skip_path()
```
saartochner, line 48: as written before, this `if` should be inside the `lumigo_dumps` function and not here.

src/lumigo_tracer/wrappers/http/http_data_classes.py
saartochner: It shouldn't be here 😂  this is not part of the `HttpRequest` 🤷 
WDYT about putting it in the `HttpState`?

src/lumigo_tracer/wrappers/http/sync_http_wrappers.py
```diff
@@ -109,7 +109,11 @@ def _update_request_data_increased_size_limit(http_info: dict, max_size: int) ->
         return
     http_info["request"].update(
         {
-            "body": lumigo_dumps(HttpState.previous_request.body, max_size)
+            "body": lumigo_dumps(
+                HttpState.previous_request.body,
+                max_size,
+                omit_skip_path=HttpState.previous_request.omit_skip_path,
```
saartochner, line 115: if you'll put the `omit_skip_path` on the state, then this line will make much more sense! :)

src/test/unit/test_lumigo_utils.py
```diff
@@ -247,12 +248,50 @@ def func():
...
+            ["R", "o", "key"],
+            '{"R": [{"o": {"key": "value"}}]}',
+        ),
+        (  # Doesnt affect other paths
+            {"a": {"key": "v"}, "b": {"key": "v"}},
+            ["a", "key"],
+            '{"a": {"key": "v"}, "b": {"key": "****"}}',
+        ),
+        (  # Nested items not affected
+            {"key": {"password": "v"}},
```
saartochner, line 280: this is surprising!
I really thought that this is part of the logic we want- think about putting a nested key on ddb. IMO, we don't want to scrub it...

src/test/unit/test_lumigo_utils.py
```diff
@@ -247,12 +248,50 @@ def func():
...
+            ["key"],
+            '{"key": {"password": "****"}}',
+        ),
+        (  # Skipping nested item should skip also parent item
```
saartochner, line 284: the comment is wrong, right?

src/lumigo_tracer/lumigo_utils.py
```diff
@@ -547,21 +549,38 @@ def _recursive_omitting(
     d, free_space = prev_result
     if free_space < 0:
         return d, free_space
+    should_skip_key = False
+    if isinstance(d, list):
+        d.append(None)
+        key = -1
+    elif omit_skip_path:
+        should_skip_key = omit_skip_path[0] == key
+        omit_skip_path = omit_skip_path[1:] if len(omit_skip_path) > 1 and should_skip_key else None
```
saartochner, line 558:
```suggestion
    elif omit_skip_path:
        should_skip_key = omit_skip_path[0] == key
        omit_skip_path = omit_skip_path[1:] if should_skip_key else None
```
Is more elegant IMO :)
(empty lists are falsy just like None)

src/test/unit/test_lumigo_utils.py
```diff
@@ -247,12 +248,50 @@ def func():
...
+            {"key": {"password": "v"}},
+            ["key", "password"],
+            '{"key": {"password": "v"}}',
+        ),
```
saartochner, line 288: There is another test that you should add:
`{"a": {"key": "c"}}, ["key"], {"a": {"key": "c"}}`
Which is: you need that the whole path will match, and not only a suffix of it.

saartochner (changes requested):
Well done!
Very cool PR! :)

src/test/unit/test_lumigo_utils.py
nirLumigo: But if someones saves passwords in DDB, we want to scrub the passwords, isn't it? 🤔

src/test/unit/test_lumigo_utils.py
nirLumigo: No, why should it be wrong?
Let's say we have the case in the test, we don't wan't to return `{"key": "****"}`

src/test/unit/test_lumigo_utils.py
saartochner: but this is inside the `Key` key, that shouldn't be scrubbed 🤔

src/test/unit/test_lumigo_utils.py
saartochner: oh I see
The comment just confused me 🤔 
Maybe the comment should be: happy flow - nested case?

src/lumigo_tracer/lumigo_utils.py
nirLumigo: I was also thinking about it but I think it might be confusing because `[]` might be interpreted as "No more items in `omit_skip_path` so we arrived to the place that should be skipped" or in other words: "when the list is empty, we walked all the path, arrived at the destination and it's time to scrub".

saartochner: yeah, but that's not how the function works 🤔 
If you'll give an empty list to the function - what should it return?

src/test/unit/test_lumigo_utils.py
nirLumigo: We just don't want scrub the immediate value of `Key` because `Key` is part of the AWS api.
We want to keep scrubbing the data of the client but not the AWS api.

src/lumigo_tracer/lumigo_utils.py
nirLumigo: But `NewType` must be mapped only to one type:
image

saartochner: `Union[List, Dict]` is a single type 🤔

src/test/unit/test_lumigo_utils.py
saartochner: I would expect that it will be: `'{"key": "****"}'`

src/lumigo_tracer/lumigo_utils.py
```diff
@@ -547,34 +553,51 @@ def _recursive_omitting(
...
+    if isinstance(d, dict) and omit_skip_path:
+        should_skip_key = omit_skip_path[0] == key
+        omit_skip_path = omit_skip_path[1:] if should_skip_key else None
     if key in SKIP_SCRUBBING_KEYS:
-        d[key] = value
-        free_space -= len(value) if isinstance(value, str) else len(aws_dump(d))
-    elif isinstance(key, str) and regex and regex.match(key):
-        d[key] = "****"
+        new_value = value
+        free_space -= len(value) if isinstance(value, str) else len(aws_dump({key: value}))
```
saartochner, line 562:
```suggestion
        free_space -= len(value) if isinstance(value, str) else len(aws_dump(d))
```
If `d` was list then this is wierd

src/lumigo_tracer/wrappers/http/http_data_classes.py
```diff
@@ -26,6 +26,7 @@ def clone(self, **kwargs):
 class HttpState:
     previous_request: Optional[HttpRequest] = None
     previous_response_body: bytes = b""
+    omit_skip_path: Optional[List[str]] = None
```
saartochner, line 29:
```suggestion
    previous_omit_skip_path: Optional[List[str]] = None
```

saartochner (approved):

PR #175, opened 2021-01-04
RD-4296 - mimic AWS import strategy of the original handler

saartochner:


src/lumigo_tracer/auto_instrument_handler.py
```diff
@@ -10,13 +10,13 @@ def parse_handler():
...
     except KeyError:
-        raise ValueError(
+        raise Exception(
             "Could not find the original handler. Please contact Lumigo for more information."
-        )
-    except ValueError:
-        raise RuntimeError(
-            f"Invalid handler format: Bad handler '{os.environ[ORIGINAL_HANDLER_KEY]}': not enough values to unpack (expected 2, got 1)"
-        )
+        ) from None
```
nirLumigo, line 15: Why are you adding `from None`? The user will lose the context of the exception:
image

Same regarding the other `from None` is this file

src/lumigo_tracer/auto_instrument_handler.py
```diff
@@ -26,11 +26,21 @@ def _handler(*args, **kwargs):
...
     try:
         handler_module, unit_name = parse_handler()
-        original_handler = getattr(importlib.import_module(handler_module), unit_name)
-    except (ImportError, AttributeError):
+        original_module = importlib.import_module(handler_module)
+    except ImportError as e:
         raise ImportError(
-            f"Unable to import module '{handler_module}': No module named '{handler_module}'"
-        )
+            f"Runtime.ImportModuleError: Unable to import module '{handler_module}': {str(e)}"
```
nirLumigo, line 32: Here you even add of original exception in the text. Don't you prefer to do `from e` instead?

src/test/unit/test_auto_instrument_handler.py
```diff
@@ -32,33 +32,67 @@ def test_hierarchy_happy_flow(monkeypatch):
 def test_import_error(monkeypatch):
     monkeypatch.setenv(ORIGINAL_HANDLER_KEY, "blabla.not.exists")
 
-    with pytest.raises(ImportError):
```
nirLumigo, line 35: Why not do it with `pytest.raises`?
Same issue regarding the other tests in this file as well.

nirLumigo (changes requested):
Nice :)

src/lumigo_tracer/auto_instrument_handler.py
saartochner: Yeah, That's what AWS does, and I want to have the same behavior...

src/lumigo_tracer/auto_instrument_handler.py
saartochner: I wanted to copy AWS's bootstrap default behavior: 
image

src/test/unit/test_auto_instrument_handler.py
saartochner: That's a great question!
I did use pytest.raises in the first place, but then I noticed that it removes the context of the message, and I want to validate that there was no context (to follow AWS behavior!) :D

src/lumigo_tracer/auto_instrument_handler.py
saartochner: Yep, I tried to copied AWS behavior:
image

src/test/unit/test_auto_instrument_handler.py
nirLumigo: Add a comment explaining why don't we use it

nirLumigo (approved):

PR #181, opened 2021-02-15
kinesis trigger by metadata 

uri-p:


src/lumigo_tracer/event/event_trigger.py
```diff
@@ -197,6 +197,10 @@ def _parse_streams(event: dict) -> Dict[str, str]:
         result.update(_parse_sqs_event(event))
     elif triggered_by == "kinesis":
         result["messageId"] = safe_get(event, ["Records", 0, "kinesis", "sequenceNumber"])
+        result["recordsNum"] = len(safe_get(event, ["Records"]))
```
efimk-lu, line 200: @saartochner in case the list is very long (10K) will `len` performance be affected?

src/lumigo_tracer/event/event_trigger.py
```diff
@@ -197,6 +197,10 @@ def _parse_streams(event: dict) -> Dict[str, str]:
...
+        event_id = safe_get(event, ["Records", 0, "eventID"])
+        if event_id:
+            result["shardId"] = event_id.split(":")[0]
```
efimk-lu, line 203: @uri-p do we index it in explore?

efimk-lu (changes requested):
See my comments

src/lumigo_tracer/event/event_trigger.py
saartochner: in python, list's `len` is O(1)

saartochner, line 200: to spare performance, you can do
```
records = safe_get(event, ["Records"])
```

src/lumigo_tracer/event/event_trigger.py
```diff
@@ -197,6 +197,10 @@ def _parse_streams(event: dict) -> Dict[str, str]:
         result.update(_parse_sqs_event(event))
     elif triggered_by == "kinesis":
         result["messageId"] = safe_get(event, ["Records", 0, "kinesis", "sequenceNumber"])
+        result["recordsNum"] = len(safe_get(event, ["Records"]))
+        event_id = safe_get(event, ["Records", 0, "eventID"])
+        if event_id:
```
saartochner, line 202:
```suggestion
        if isinstance(event_id, str):
```

saartochner (changes requested):

src/lumigo_tracer/event/event_trigger.py
uri-p: Not by default, but I do want to add it in the future as an execution tag that gets index.
It should be the tradeoff between reducing the size of the event (those metadata fields are duplicates) and add those metadata fields as execution tags.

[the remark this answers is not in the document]
uri-p: What the meaning of this change ?

src/lumigo_tracer/event/event_trigger.py
uri-p: Not sure I understand what is the improvement here?

saartochner: checkout lines 199-201 - they all do the same thing

uri-p: I don't follow you, all those safe_get are o(1) complexity no ?

src/lumigo_tracer/event/event_trigger.py
efimk-lu: So where exactly is it written now?

src/lumigo_tracer/event/event_trigger.py
saartochner: yes.
But they are wrapped with many `try/except` which may cause performance issues on the edgecase of many small requests.
Well. It is not that important. I'm taking it back. 👍

src/lumigo_tracer/event/event_trigger.py
uri-p: They will be part of the trigger by information.
I want to add auto-tagging for shard id and we need to think about how to make it easy to use the bulk size.

You will be able at first to see the distribution of your invocation and shards, make it easy to understand that maybe your partition key is not optimized, currently you have to enable CW per shard metrics to find this information. 
This can be done with lumigo OOB

efimk-lu: So you'll add it to get transaction return value as well, right?

efimk-lu (approved):

src/lumigo_tracer/event/event_trigger.py
uri-p: It can be done, once we will decide who to present it in the UI we can easily expose it

PR #187, opened 2021-04-12
find s3 bucket name from uri (and not just from host)

sapirlumigo:


src/lumigo_tracer/wrappers/http/http_parser.py
```diff
@@ -318,7 +321,7 @@ def get_parser(url: str, headers: Optional[dict] = None) -> Type[Parser]:
         return KinesisParser
     elif service == "events":
         return EventBridgeParser
-    elif safe_split_get(url, ".", 1) == "s3":
+    elif safe_split_get(url, ".", 1) == "s3" or safe_split_get(url, ".", 0) == "s3":
```
nirLumigo, line 324: Add a test for this so no one will accidentally break it

nirLumigo (changes requested):
Nice! :)

nirLumigo (approved):

PR #191, opened 2021-05-10
RD-4996 added memory over time to singe transaction

OrrLumigo:


src/lumigo_tracer/extension/extension.py
```diff
@@ -48,7 +52,8 @@ def shutdown(self):
             self._finish_previous_invocation(current_bandwidth)
 
     def _finish_previous_invocation(self, current_bandwidth: Optional[int]):
-        self.sampler.stop_sampling()
```
saartochner, line 51: there should be a single sampler.
Maybe you can change this sampler to collect two metrics!

src/lumigo_tracer/extension/extension.py
```diff
@@ -69,6 +75,7 @@ def _finish_previous_invocation(self, current_bandwidth: Optional[int]):
             started=int(self.start_time.timestamp() * 1000),
             requestId=self.request_id,
             networkBytesUsed=current_bandwidth - self.bandwidth,
-            cpuUsageTime=[s.dump() for s in self.sampler.get_samples()],
+            cpuUsageTime=[s.dump() for s in self.spu_sampler.get_samples()],
+            memoryUsageTime=[s.dump() for s in self.memory_sampler.get_samples()],
```
saartochner, line 79:
```suggestion
            cpuUsageTime=[s.dump() for s in self.sampler.get_cpu_samples()],
            memoryUsageTime=[s.dump() for s in self.sampler.get_memory_samples()],
```
This is what I would have expect

saartochner (changes requested):
GJ!

src/lumigo_tracer/extension/extension_utils.py
```diff
@@ -20,6 +20,14 @@ def get_current_cpu_time() -> Optional[int]:
         return total
 
 
+def get_current_memory() -> Optional[int]:
+    """
+    :return: the total number of milliseconds that being used by the CPU.
+    """
+    with lumigo_safe_execute("Extension: get cpu time"):
```
saartochner, line 27:
```suggestion
    with lumigo_safe_execute("Extension: get memory usage"):
```

src/test/unit/extension/conftest.py
```diff
@@ -21,7 +21,18 @@ def sampler_timer(monkeypatch):
 
 
 @pytest.fixture
-def mock_linux_files():
```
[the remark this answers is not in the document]
OrrLumigo: but this return a const value for readlines..

src/lumigo_tracer/extension/extension_utils.py
```diff
@@ -20,6 +21,19 @@ def get_current_cpu_time() -> Optional[int]:
         return total
 
 
+def get_current_memory() -> Optional[float]:
+    with lumigo_safe_execute("Extension: get meminfo"):
+        with open("/proc/meminfo", "r") as meminfo:
+            meminfo_content = meminfo.read()
+            mem_available_pattern = re.compile(r'(MemAvailable)[:][ ]*([0-9]*)')
+            mem_total_pattern = re.compile(r'(MemTotal)[:][ ]*([0-9]*)')
```
saartochner, line 29: put it as a constant . you don't need to compile it in every sample

saartochner (approved):

PR #193, opened 2021-06-14
log fatal error: print message if could not send span

saartochner:


doriaviram (approved):

saartochner:
created another with the same code

saartochner:
created another with new name

PR #203, opened 2021-07-13
requests api should be traced as a whole

saartochner:


src/lumigo_tracer/wrappers/http/sync_http_wrappers.py
```diff
@@ -204,9 +205,28 @@ def _requests_wrapper(func, instance, args, kwargs):
...
+                span = add_request_event(
+                    HttpRequest(
+                        host=url,
+                        method=args[0].upper(),
+                        uri=url,
+                        body=kwargs.get("data"),
+                        headers=kwargs.get("headers"),
+                    )
+                )
+            SpansContainer.get_span().add_exception_to_span(span, exception, [])
```
nirLumigo, line 228: Won't it add the exception for the entire execution?
I think the error suppose to be added only to the http call.

src/test/unit/wrappers/http/test_sync_http_wrappers.py
```diff
@@ -265,6 +265,35 @@ def delayed_getaddrinfo(*args, **kwargs):
...
+    @lumigo_tracer.lumigo_tracer()
+    def lambda_test_function(event, context):
+        try:
+            requests.post("https://www.google.com", data=b"123", headers={"a": "b"})
+        except ZeroDivisionError:
+            return True
+        return False
+
+    # requests executes this function before/after the http call
+    monkeypatch.setattr(*func_to_patch, lambda *args, **kwargs: 1 / 0)
```
nirLumigo, line 285: 💪

nirLumigo (changes requested):
Nice! :)

src/lumigo_tracer/wrappers/http/sync_http_wrappers.py
saartochner: no, it'll add exception to the provided span. This span is th http span (and not the function span)

nirLumigo: This might be problematic:
- The user might wrap the call in try-except so he might not be interested in this error but we'll send him an alert.
- What happens if there is a real exception in the lambda? Won't it override the error?
- What happens there are multiple http calls that fails? Each one will override the previous one?

src/lumigo_tracer/wrappers/http/sync_http_wrappers.py
```diff
@@ -204,9 +205,28 @@ def _requests_wrapper(func, instance, args, kwargs):
...
+            if HttpState.previous_request and HttpState.previous_request.host in url:
+                span = SpansContainer.get_span().get_last_span()
+            else:
+                span = add_request_event(
+                    HttpRequest(
+                        host=url,
+                        method=args[0].upper(),
+                        uri=url,
+                        body=kwargs.get("data"),
+                        headers=kwargs.get("headers"),
```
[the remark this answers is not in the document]
saartochner: nope :)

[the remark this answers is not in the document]
saartochner: image

src/lumigo_tracer/wrappers/http/sync_http_wrappers.py
saartochner: after the talk changed to static method to avoid future confusion

nirLumigo (commented):
GJ :)

nirLumigo (approved):

PR #208, opened 2021-07-27
support static ip reporter

uri-p:


src/test/unit/test_lumigo_utils.py
```diff
@@ -464,14 +469,15 @@ def asserting_extension(file_path, single):
...
 )
-def test_report_json_retry(monkeypatch, reporter_mock, caplog, errors, final_log):
+def test_report_json_retry(monkeypatch, reporter_mock, caplog, errors, final_log, host):
     reporter_mock.side_effect = report_json
-    monkeypatch.setattr(Configuration, "host", "force_reconnect")
+    monkeypatch.setattr(Configuration, "host", host)
     monkeypatch.setattr(Configuration, "should_report", True)
     monkeypatch.setattr(http.client, "HTTPSConnection", Mock())
-    http.client.HTTPSConnection("force_reconnect").getresponse.side_effect = errors
+    http.client.HTTPSConnection(host).getresponse.side_effect = errors
```
saartochner, line 480: Not sure what you did here.
This test checks the retry mechanism in `report_json` - 
I.e. catch an error on the first try and perform retry.

I think that you need to create another test if you want to test your behavior

src/lumigo_tracer/lumigo_utils.py
```diff
@@ -277,12 +279,25 @@ def _create_request_body(
...
+                host, timeout=EDGE_TIMEOUT, context=ssl._create_unverified_context()
+            )
+        else:
+            return http.client.HTTPSConnection(host, timeout=EDGE_TIMEOUT)
     except Exception as e:
         get_logger().exception(f"Could not establish connection to {host}", exc_info=e)
     return None
 
 
+def is_valid_ip(host: str) -> bool:
```
saartochner, line 293: the name is a bit confusing. This function only checks that the input looks like an IP, 
maybe we can call it "is_ip_address"?

src/lumigo_tracer/lumigo_utils.py
```diff
@@ -277,12 +279,25 @@ def _create_request_body(
 
 def establish_connection(host):
     try:
-        return http.client.HTTPSConnection(host, timeout=EDGE_TIMEOUT)
+        if is_valid_ip(host):
+            return http.client.HTTPSConnection(
+                host, timeout=EDGE_TIMEOUT, context=ssl._create_unverified_context()
+            )
+        else:
+            return http.client.HTTPSConnection(host, timeout=EDGE_TIMEOUT)
```
saartochner, line 287: try to reduce duplication:
```
context = ssl._create_unverified_context() if is_ip_address(host) else None
return http.client.HTTPSConnection(host, timeout=EDGE_TIMEOUT, context=context)
```
(Not mandatory)

saartochner (changes requested):
Cool!!

PR #209, opened 2021-08-02
add connection id

uri-p:


src/lumigo_tracer/spans_container.py
```diff
@@ -73,6 +74,7 @@ def __init__(
         self.transaction_id = transaction_id
         self.max_finish_time = max_finish_time
         self.base_msg = {
+            "lambda_container_id": SpansContainer.lambda_container_id,
```
saartochner, line 77: do you want it in the HTTP spans as well?
If no, you should put it under seelf.function_span

src/lumigo_tracer/wrappers/http/http_data_classes.py
```diff
@@ -8,13 +8,15 @@ class HttpRequest:
...
     body: bytes
+    instance_id: bytes
 
     def __init__(self, **kwargs):
         self.host = kwargs["host"]
         self.method = kwargs["method"]
         self.uri = kwargs["uri"]
         self.headers = {k.lower(): v for k, v in (kwargs.get("headers") or {}).items()}
         self.body = kwargs.get("body")
+        self.instance_id = kwargs.get("instance_id")
```
saartochner, line 19: self.instance_id maybe should be `Optional[bytes]`?

src/lumigo_tracer/wrappers/http/sync_http_wrappers.py
```diff
@@ -175,10 +175,19 @@ def _http_send_wrapper(func, instance, args, kwargs):
...
+                    uri=uri,
+                    headers=headers,
+                    body=body,
+                    instance_id=id(instance),
+                )
             )
         else:
-            add_unparsed_request(HttpRequest(host=host, method=method, uri=uri, body=data))
+            add_unparsed_request(
+                HttpRequest(host=host, method=method, uri=uri, body=data, instance_id=id(instance))
```
saartochner, line 189: what is an `id` of an `instance` here?
This is an instance of an `HTTPConnection` which doesn't tell you much about the request/session/machine 🤔 
What are you trying to achieve?

saartochner (changes requested):
GJ!

src/lumigo_tracer/spans_container.py
uri-p: Yes for all connections

src/lumigo_tracer/wrappers/http/sync_http_wrappers.py
uri-p: The goal is to validate the reuse of connections.
For example:
Multiple connections to the same AWS service like kinesis lambda etc.

The data is based on HTTP/S connection id + host

src/lumigo_tracer/wrappers/http/http_data_classes.py
uri-p: Is it the same as 'body'?
I don't see any optional

saartochner: You right, change body too :)

saartochner (approved):

PR #211, opened 2021-08-09
change importlib to imp.load_module

GuyMoses:


src/lumigo_tracer/auto_instrument_handler.py
```diff
@@ -1,3 +1,4 @@
+import imp
```
saartochner, line 1: you need to catch the warning and suppress it (just like was did)

src/lumigo_tracer/test_module/test.py
```diff
@@ -0,0 +1,2 @@
+def cde(*args, **kwargs):
```
saartochner, line 1: remove

doriaviram: He using that on 
```monkeypatch.setenv(ORIGINAL_HANDLER_KEY, "lumigo_tracer/test_module/test.cde")```

saartochner, line 1: chnge the name to `handler` 😂

saartochner (approved):
Great work!

PR #212, opened 2021-08-09
RD-4775 - support threading in tracer

saartochner:


src/lumigo_tracer/wrappers/http/sync_http_wrappers.py
```diff
@@ -224,7 +231,7 @@ def _requests_wrapper(func, instance, args, kwargs):
             method = safe_get_list(args, 0, kwargs.get("method", "")).upper()
             url = safe_get_list(args, 1, kwargs.get("url"))
             if HttpState.previous_request and HttpState.previous_request.host in url:
-                span = SpansContainer.get_span().get_last_span()
+                span = SpansContainer.get_span().get_span_by_id(HttpState.previous_span_id)
```
doriaviram, line 234: It's not the same as it was before? taking the last span that was sent and not the current one?

saartochner: This is the only case that we're using this "last one" - this is relevant only to an extremely specific case wherein the `requests` package failed in some outer-processing logic (i.e. not part of the HTTP request).
In this case, I don't have any other id that can redirect me to the request itself (instance is None), so I'm using a heuristic of validating the URL 🤷

src/lumigo_tracer/examples/threads.py
```diff
@@ -0,0 +1,25 @@
+import concurrent.futures
```
nirLumigo, line 1: Why do we need the example file? Looks like the user doesn't have to do anything special in order to use threads.

src/lumigo_tracer/spans_container.py
```diff
@@ -150,44 +150,47 @@ def add_span(self, span: dict) -> dict:
         This function parses an request event and add it to the span.
         """
         new_span = recursive_json_join(span, self.base_msg)
-        self.spans.append(new_span)
-        self.span_ids_to_send.add(span["id"])
+        span_id = new_span["id"]
+        self.spans[span_id] = new_span
+        self.span_ids_to_send.add(span_id)
```
nirLumigo, line 155: Why do we need `span_ids_to_send`? Maybe we don't need it anymore?

src/lumigo_tracer/spans_container.py
```diff
@@ -150,44 +150,47 @@ def add_span(self, span: dict) -> dict:
...
         return new_span
 
-    def get_last_span(self) -> Optional[dict]:
-        if not self.spans:
+    def get_span_by_id(self, span_id: Optional[str]) -> Optional[dict]:
+        if not span_id:
             return None
```
nirLumigo, line 160: I don't think that we need those lines, you use `dict.get`

src/lumigo_tracer/spans_container.py
```diff
@@ -150,44 +150,47 @@ def add_span(self, span: dict) -> dict:
...
-        for span in self.spans:
-            if span.get("id") == span_id:
-                return span
-        return None
-
-    def pop_last_span(self) -> Optional[dict]:
-        return self.spans.pop() if self.spans else None
+    def pop_span(self, span_id: Optional[str]) -> Optional[dict]:
+        if not span_id:
+            return None
```
nirLumigo, line 165: I don't think that we need those lines, you use `dict.pop` with default value

src/lumigo_tracer/wrappers/http/http_data_classes.py
```diff
@@ -27,10 +27,13 @@ def clone(self, **kwargs):
 
 class HttpState:
     previous_request: Optional[HttpRequest] = None
-    previous_response_body: bytes = b""
+    previous_span_id: Optional[str] = None
     omit_skip_path: Optional[List[str]] = None
+    request_to_span_id: Dict[int, str] = {}
```
nirLumigo, line 32:
```suggestion
    span_id_to_request: Dict[int, str] = {}
```
Isn't it?

src/lumigo_tracer/wrappers/http/http_data_classes.py
```diff
@@ -27,10 +27,13 @@ def clone(self, **kwargs):
...
+    response_to_span_id: Dict[int, str] = {}
```
nirLumigo, line 33:
```suggestion
    span_id_to_response: Dict[int, str] = {}
```
Isn't it?

src/lumigo_tracer/wrappers/http/sync_http_wrappers.py
```diff
@@ -36,20 +38,23 @@ def is_lumigo_edge(host: Optional[str]) -> bool:
     return False
 
 
-def add_request_event(parse_params: HttpRequest) -> Dict:
+def add_request_event(span_id: Optional[str], parse_params: HttpRequest) -> Dict:
```
nirLumigo, line 41: Why is it `Optional`? Should it always exist?

src/lumigo_tracer/wrappers/http/sync_http_wrappers.py
```diff
@@ -58,23 +63,22 @@ def add_unparsed_request(parse_params: HttpRequest):
...
                 if "response" not in last_event["info"]["httpInfo"]:
-                    SpansContainer.get_span().pop_last_span()
+                    SpansContainer.get_span().pop_span(span_id)
                     body = (HttpState.previous_request.body + parse_params.body)[
                         : get_size_upper_bound()
                     ]
-                    add_request_event(HttpState.previous_request.clone(body=body))
-                    return
-    add_request_event(parse_params.clone(headers=None))
+                    return add_request_event(span_id, HttpState.previous_request.clone(body=body))
```
nirLumigo, line 76: Shouldn't we stop using `previous_request`?

src/lumigo_tracer/wrappers/http/sync_http_wrappers.py
```diff
@@ -83,25 +87,23 @@ def update_event_response(
...
         return
-    last_event = SpansContainer.get_span().pop_last_span()
+    last_event = SpansContainer.get_span().pop_span(span_id)
     if last_event:
         http_info = last_event.get("info", {}).get("httpInfo", {})
         if not host:
             host = http_info.get("host", "unknown")
-        else:
-            HttpState.previous_response_body = b""
+        body = http_info.get("response", {}).get("body", "").encode() + body
```
nirLumigo, line 95: Looks like now we always concat the `body` but there should be ` limit, looks like you deleted this part
image

src/lumigo_tracer/wrappers/http/sync_http_wrappers.py
```diff
@@ -122,9 +124,29 @@ def _update_request_data_increased_size_limit(http_info: dict, max_size: int) ->
     )
 
 
+def get_lumigo_connection_id(instance) -> Optional[int]:
+    return getattr(instance, LUMIGO_CONNECTION_ID_KEY, None)
+
+
+def set_lumigo_connection_id(instance):
+    r = random.random()
+    print("Choose random", r)
```
nirLumigo, line 133: Won't it print to the Cloudwatch of the customer?

src/lumigo_tracer/wrappers/http/sync_http_wrappers.py
```diff
@@ -172,26 +194,33 @@ def _http_send_wrapper(func, instance, args, kwargs):
...
         else:
-            add_unparsed_request(
-                HttpRequest(host=host, method=method, uri=uri, body=data, instance_id=id(instance))
+            span = add_unparsed_request(
+                HttpState.request_to_span_id.get(get_lumigo_connection_id(instance)),
+                HttpRequest(host=host, method=method, uri=uri, body=data, instance_id=id(instance)),
             )
+        span_id = span["id"] if span else None
+        if span_id:
+            HttpState.request_to_span_id[get_lumigo_connection_id(instance)] = span_id
```
nirLumigo, line 218:
```suggestion
            HttpState.connection_id_to_span_id[get_lumigo_connection_id(instance)] = span_id
```
Isn't it?

src/lumigo_tracer/wrappers/http/sync_http_wrappers.py
doriaviram: I think in this case the best is to send "half" span, this is just a mistake and the user can't notice that

nirLumigo: Don't you have `instance` so you should have the id you attached to the `HTTPConnection` object?

src/lumigo_tracer/wrappers/http/sync_http_wrappers.py
```diff
@@ -122,9 +124,29 @@ def _update_request_data_increased_size_limit(http_info: dict, max_size: int) ->
...
+def set_lumigo_connection_id(instance):
+    r = random.random()
+    print("Choose random", r)
+    setattr(instance, LUMIGO_CONNECTION_ID_KEY, r)
+
+
 #   Wrappers  #
 
 
+def _http_init_wrapper(func, instance, args, kwargs):
```
nirLumigo, line 140: Won't we have an issue with sync calls that use the same `HTTPConnection` object?

src/lumigo_tracer/wrappers/sql/sqlalchemy_wrapper.py
```diff
@@ -40,7 +45,7 @@ def _before_cursor_execute(conn, cursor, statement, parameters, context, execute
 
 def _after_cursor_execute(conn, cursor, statement, parameters, context, executemany):
     with lumigo_safe_execute("handle sqlalchemy after execute"):
-        span = SpansContainer.get_span().get_last_span()
+        span = SpansContainer.get_span().get_span_by_id(_last_span_id)
```
nirLumigo, line 48: In this file, I think we still take only the last_span. I'm not sure it will always work with threads.

src/test/unit/wrappers/http/test_sync_http_wrappers_threads.py
```diff
@@ -0,0 +1,31 @@
+import urllib.request
```
nirLumigo, line 1: Usually the name of the test file is the same as the name of the real file

nirLumigo (changes requested):
Good job! hard task 💪

src/lumigo_tracer/spans_container.py
saartochner: We're using it in order to not send spans twice, given the timeout mechanism:
It is relevant only to the case that the TO mechanism has been executed, and then the lambda *ended normally*
So we don't want to send the same spans twice :)

src/lumigo_tracer/wrappers/http/sync_http_wrappers.py
saartochner: nope. Sometimes you want to override a span (e.g. when you download a stream, and you get another part from the body), and sometimes you want to create a new span.

src/lumigo_tracer/wrappers/http/sync_http_wrappers.py
saartochner: nope, this part exists inside `parser.parse_response`. It was here because we used to save the body in two places (both the span, and inside the HttpState), and now we save it only in the span :)

src/lumigo_tracer/wrappers/http/sync_http_wrappers.py
saartochner: my bad sorry!!!!!

src/lumigo_tracer/wrappers/sql/sqlalchemy_wrapper.py
saartochner: yes. I preferred to not support sqlalchemy with threads. this is too niche

src/lumigo_tracer/spans_container.py
saartochner: mypy otherwise fail

src/lumigo_tracer/spans_container.py
saartochner: mypy otherwise fail

nirLumigo (changes requested):

src/lumigo_tracer/wrappers/http/sync_http_wrappers.py
```diff
@@ -66,14 +66,14 @@ def add_unparsed_request(span_id: Optional[str], parse_params: HttpRequest) -> O
...
-                        : get_size_upper_bound()
-                    ]
-                    return add_request_event(span_id, HttpState.previous_request.clone(body=body))
+        if last_event and last_event.get("type") == HTTP_TYPE:
+            http_info = last_event.get("info", {}).get("httpInfo", {})
+            if http_info.get("host") == parse_params.host:
+                if "response" not in http_info:
+                    SpansContainer.get_span().get_span_by_id(span_id)
+                    prev_body = http_info.get("request", {}).get("body", '""').encode()[1:-1]
+                    http_info["request"]["body"] = lumigo_dumps(prev_body + parse_params.body)
```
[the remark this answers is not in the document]
saartochner: It's a string ,and I'm using `lumigo_dump` mainly to limit size

src/test/unit/wrappers/http/test_sync_http_wrappers.py
```diff
@@ -552,17 +552,13 @@ def lambda_test_function(event, context):
...
+    response_no_error = http_info_no_error["response"]
 
     assert http_info_no_error["response"]["statusCode"] == 200
     assert http_info_with_error["response"]["statusCode"] >= 400
 
-    assert len(request_with_error["headers"]) > len(request_no_error["headers"])
-    assert request_with_error["headers"] == json.dumps(d)
-
-    assert len(request_with_error["body"]) > len(request_no_error["body"])
-    assert request_with_error["body"] == json.dumps(d)
```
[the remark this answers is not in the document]
saartochner: you're right. good catch :)

nirLumigo (changes requested):
Nice :)

src/lumigo_tracer/wrappers/http/sync_http_wrappers.py
```diff
@@ -71,10 +72,13 @@ def add_unparsed_request(span_id: Optional[str], parse_params: HttpRequest) -> O
             if http_info.get("host") == parse_params.host:
                 if "response" not in http_info:
                     SpansContainer.get_span().get_span_by_id(span_id)
-                    prev_body = http_info.get("request", {}).get("body", '""').encode()[1:-1]
+                    # We're stripping the prev body from the previous `lumigo_dumps`
+                    prev_body = http_info.get("request", {}).get("body", "").encode().strip(b'"')
                     http_info["request"]["body"] = lumigo_dumps(prev_body + parse_params.body)
+                    if HttpState.previous_span_id == span_id and HttpState.previous_request:
+                        HttpState.previous_request.body = prev_body + parse_params.body
```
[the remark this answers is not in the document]
saartochner: yes. 
you're getting the next part of the HTTP body :)

doriaviram (changes requested):
Remove 
`span = SpansContainer.get_span().get_span_by_id(HttpState.previous_span_id)`

src/lumigo_tracer/lumigo_utils.py
```diff
@@ -747,6 +747,22 @@ def lumigo_dumps(
     )
 
 
+def concat_old_body_to_new(old_body: Optional[str], new_body: bytes) -> str:
+    """
+    We have only a dumped body from the previous request,
+    so to concatenate the new body we should undo the lumigo_dumps.
+    Note that the old body is dumped bytes
```
nirLumigo, line 754: But the type is `old_body: Optional[str]` and not `old_body: Optional[bytes]`

src/test/unit/test_lumigo_utils.py
```diff
@@ -625,3 +626,18 @@ def test_internal_analytics_message(capsys):
...
+
+
+@pytest.mark.parametrize(
+    "old, new, expected",
+    [
+        (b"1", b"2", b"12"),  # happy flow
+        (b"", b"2", b"2"),  # no old
+        (b"1", b"", b"1"),  # no new
+        (b"12", b"34567", b"12345...[too long]"),  # together pass max size
+        (b"123456", b"789", b"12345...[too long]"),  # old pass max size
```
nirLumigo, line 638: Add a test with json and in the json there's a string with the `"` inside the string.
Something like: `json.dumps({"k": 'h"ell"o'})`

nirLumigo (changes requested):
Nice :)

src/lumigo_tracer/lumigo_utils.py
saartochner: you right, dumping changes the data to str.
But dumping could fail, so this is Optional...

nirLumigo (approved):

src/lumigo_tracer/wrappers/http/sync_http_wrappers.py
saartochner: ok, we will do this `HttpState.previous_span_id` only if we have a flag of IS_TRACER_SYNC 👌

doriaviram (approved):

PR #213, opened 2021-08-17
RD-5176 - don't retry to send after timeout to edge

saartochner:


src/lumigo_tracer/lumigo_utils.py
```diff
@@ -106,6 +106,16 @@ def get_region() -> str:
     pass
 
 
+class InternalState:
+    timeout_on_connection = False
+    internal_error_already_logged = False
+
+    @staticmethod
+    def reset():
```
nirLumigo, line 114: You call `reset` only in conftest. Don't you need to call it also at the beginning of the tracer execution? Won't it leak when there's a warm start?

nirLumigo (changes requested):
Nice :)

src/lumigo_tracer/lumigo_utils.py
saartochner: It should leak. This is a global state that we should have :)

saartochner: E.g. - when there is a timeout to the edge, don't waste more time on it (even in future invocations)

src/lumigo_tracer/lumigo_utils.py
```diff
@@ -107,14 +109,25 @@ def get_region() -> str:
...
-        InternalState.timeout_on_connection = False
+        InternalState.timeout_on_connection = None
         InternalState.internal_error_already_logged = False
 
+    @staticmethod
+    def mark_timeout_to_edge():
+        InternalState.timeout_on_connection = datetime.datetime.now()
+
+    @staticmethod
+    def should_report_to_edge():
```
nirLumigo, line 125:
```suggestion
    def should_report_to_edge() -> bool:
```

src/lumigo_tracer/lumigo_utils.py
```diff
@@ -34,6 +35,7 @@
 HTTPS_PREFIX = "https://"
 LOG_FORMAT = "#LUMIGO# - %(asctime)s - %(levelname)s - %(message)s"
 SECONDS_TO_TIMEOUT = 0.5
+COOLDOWN_AFTER_TIMEOUT_DURATION = 10
```
nirLumigo, line 38: IMO this is easier to understand
```suggestion
COOLDOWN_AFTER_TIMEOUT_DURATION = datetime.timedelta(seconds=10)
```

nirLumigo (approved):

src/lumigo_tracer/lumigo_utils.py
```diff
@@ -753,3 +770,17 @@ def is_provision_concurrency_initialization() -> bool:
 def get_stacktrace(exception: Exception) -> str:
     original_traceback = traceback.format_tb(exception.__traceback__)
     return "".join(filter(lambda line: STACKTRACE_LINE_TO_DROP not in line, original_traceback))
+
+
+try:
```
saartochner, line 775: moved from above to reuse `establish_connection`

nirLumigo (approved):

PR #215, opened 2021-08-18
RD-5931 - follow aws logic of handler importing on auto instrument. 

saartochner:
Use the runtime's logic if possible

### DIFF
--- a/src/lumigo_tracer/libs/bootstrap.py
+++ b/src/lumigo_tracer/libs/bootstrap.py
@@ -1,0 +1,80 @@
+"""
+Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+"""
+
+import warnings
+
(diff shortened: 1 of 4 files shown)


src/lumigo_tracer/libs/bootstrap.py
```diff
@@ -0,0 +1,80 @@
+"""
```
saartochner, line 1: This file was copied from AWS's runtime

src/lumigo_tracer/libs/lambda_runtime_exception.py
```diff
@@ -0,0 +1,19 @@
+"""
```
saartochner, line 1: This file was copied from AWS's runtime

GuyMoses (approved):

nirLumigo (approved):
Great! :)

PR #217, opened 2021-08-25
Rd 5700 extension timeouts

OrrLumigo:


src/lumigo_tracer/lumigo_utils.py
```diff
@@ -379,14 +383,16 @@ def write_extension_file(data: dict, span_type: str):
...
-    get_logger().info(f"writing [{len(spans)}] spans to files, spans: {len(to_send)}")
+    get_logger().info(f"writing [{len(to_send)}] spans to files to [{EXTENSION_DIR}]")
     Path(EXTENSION_DIR).mkdir(parents=True, exist_ok=True)
     for span in to_send:
         write_extension_file(span, "span")
-    done_object = {"spansCount": len(to_send)}
-    write_extension_file(done_object, "done")
+    if with_done:
+        done_object = {"spansCount": len(to_send)}
+        print(f"created done - {done_object}")
```
nirLumigo, line 394: use `logger` instead of `print`

src/lumigo_tracer/spans_container.py
```diff
@@ -118,8 +117,10 @@ def _generate_start_span(self) -> dict:
...
         to_send = self._generate_start_span()
-        if not Configuration.send_only_if_error and not should_use_tracer_extension():
-            report_duration = lumigo_utils.report_json(region=self.region, msgs=[to_send])
+        if not Configuration.send_only_if_error:
+            report_duration = lumigo_utils.report_json(
+                region=self.region, msgs=[to_send], with_done=False
+            )
             self.function_span["reporter_rtt"] = report_duration
         else:
             get_logger().debug("Skip sending start because tracer in 'send only if error' mode .")
```
nirLumigo, line 126: Can you add a test reproducing the bug and verifying it doesn't happen anymore?

nirLumigo (changes requested):
Nice! :)

src/lumigo_tracer/lumigo_utils.py
```diff
@@ -309,13 +309,17 @@ def get_edge_host(region: Optional[str] = None) -> str:
     return host
 
 
-def report_json(region: Optional[str], msgs: List[dict], should_retry: bool = True) -> int:
+def report_json(
+    region: Optional[str], msgs: List[dict], should_retry: bool = True, with_done=True
+) -> int:
```
saartochner, line 314: I wouldn't tak it as a parameter.
If we use extension, then we most do it

saartochner: which is more about functionality and less about implementation - a better API design :)

src/test/unit/test_lumigo_utils.py
```diff
@@ -465,12 +465,12 @@ def test_report_json_extension_spans_mode(monkeypatch, reporter_mock):
...
             }
         )
-    report_json(None, spans)
+    report_json(region=None, msgs=spans, is_start_span=False)
 
     files_paths = []
     for span in spans:
         files_paths.append(get_span_file_name(span, "span"))
-    done_object = {"spansCount": len(spans)}
+    done_object = {"spansCount": len(spans) + 1}
```
saartochner, line 473: if `Configuration.send_only_if_error` then you will not send `+1` spans (because we're not sending the start span)

saartochner: please add test to this scenario :)

src/test/conftest.py
```diff
@@ -20,7 +20,9 @@
 
 
 @pytest.fixture(autouse=True)
-def reporter_mock(monkeypatch):
+def reporter_mock(monkeypatch, request):
+    if request.node.get_closest_marker("dont_mock_lumigo_utils_reporter"):
+        return
```
saartochner, line 25: whattttttttt
dude, this is something new
🚀 🔥 🥇

saartochner: just change the marker name to `skip_reporter_mock` to make it more readable :)

src/test/unit/test_lumigo_utils.py
OrrLumigo: Is there some way to know how manny spans in total exists?
i know in the node its there

saartochner: nothing that I'm aware of 🤔

PR #218, opened 2021-08-25
RD-6042 - support aiohttp

saartochner:


src/lumigo_tracer/wrappers/aiohttp/aiohttp_wrapper.py
```diff
@@ -0,0 +1,87 @@
...
+    aiohttp = None
+
+LUMIGO_SPAN_ID_KEY = "_lumigo_span_id"
+trace_config = None
+
+
+def aiohttp_session_init_wrapper(func, instance, args, kwargs):
+    with lumigo_safe_execute("aiohttp aiohttp_session_init_wrapper"):
+        traces = kwargs.get("trace_configs") or []
+        traces.append(trace_config)
```
nirLumigo, line 19: Are we sure that the type of `traces` is always a `list`?

src/lumigo_tracer/wrappers/aiohttp/aiohttp_wrapper.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+def wrap_aiohttp():
+    with lumigo_safe_execute("wrap http calls"):
+        get_logger().debug("wrapping http requests")
+        if aiohttp:
+            wrap_function_wrapper(
+                "aiohttp.client", "ClientSession.__init__", aiohttp_session_init_wrapper
+            )
+            global trace_config
```
nirLumigo, line 81: Why use a global variable and set it inside a function?
You can pass it to each function if you don't want to create it inside each function.
Also, I think it is used only in one function.

src/lumigo_tracer/wrappers/aiohttp/aiohttp_wrapper.py
```diff
@@ -0,0 +1,87 @@
...
+                body=b"",
+            ),
+        )
+        setattr(trace_config_ctx, LUMIGO_SPAN_ID_KEY, span["id"])
+
+
+async def on_request_chunk_sent(session, trace_config_ctx, params):
+    with lumigo_safe_execute("aiohttp on_request_chunk_sent"):
+        span_id = getattr(trace_config_ctx, LUMIGO_SPAN_ID_KEY)
+        span = SpansContainer.get_span().get_span_by_id(span_id)
```
nirLumigo, line 42: 💪

src/lumigo_tracer/wrappers/aiohttp/aiohttp_wrapper.py
```diff
@@ -0,0 +1,87 @@
...
+
+async def on_request_start(session, trace_config_ctx, params):
+    with lumigo_safe_execute("aiohttp on_request_start"):
+        span = add_request_event(
+            span_id=None,
+            parse_params=HttpRequest(
+                host=params.url.host,
+                method=params.method,
+                uri=str(params.url),
+                headers=dict(params.headers),
```
nirLumigo, line 32: What can it be if not a `dict`? Is it always something that can be converted to a `dict`?

src/test/unit/wrappers/http/test_sync_http_wrappers_threads.py
```diff
@@ -29,3 +32,25 @@ def lambda_test_function(event, context):
...
         assert request_index == response_index
+
+
+def test_run_in_executor(context, token):
+    @lumigo_tracer.lumigo_tracer(token=token)
+    def lambda_test_function(event, context):
+        async def main():
+            loop = asyncio.get_event_loop()
+            future1 = loop.run_in_executor(None, requests.get, "http://www.google.com")
+            response1 = await future1
```
nirLumigo, line 43: I think it's worth doing 2 API calls (with async.gather) to validate that it really works with concurrency.

nirLumigo (changes requested):
Nice! :)

src/lumigo_tracer/wrappers/aiohttp/aiohttp_wrapper.py
saartochner: That's the api, but I'll add protection 🤔

src/lumigo_tracer/wrappers/aiohttp/aiohttp_wrapper.py
saartochner: This is `CIMultiDict` from the docs, can't be something else. If it will be - I have protection :)

src/lumigo_tracer/wrappers/aiohttp/aiohttp_wrapper.py
saartochner: It is indeed used only by `aiohttp_session_init_wrapper`, but it's signature is dictated by `wrap_function_wrapper`.
I can wrap it with another function, but it might be very confusing. WDYT?

nirLumigo: I think it's better to wrap it with another function.
Global variables might hit us in places we don't expect (warm starts, unittests etc.), better to avoid stateful when possible.

nirLumigo (approved):

PR #221, opened 2021-10-05
skip parsing messageId and triggerBy when using extension

OrrLumigo:


src/test/unit/event/test_event_trigger.py
```diff
@@ -473,3 +475,14 @@
 def test_parse_triggered_by(event, output):
     Configuration.is_step_function = True
     assert parse_triggered_by(event) == output
+
+
+def test_parse_triggered_by_with_extension_should_be_unknown(monkeypatch):
+    monkeypatch.setattr(os, "environ", {"LUMIGO_USE_TRACER_EXTENSION": "TRUE"})
```
saartochner, line 481: create a fixture that does it. call it "with_extension"

src/test/conftest.py
```diff
@@ -35,6 +35,11 @@ def cancel_timeout_mechanism(monkeypatch):
     monkeypatch.setattr(Configuration, "timeout_timer", False)
 
 
+@pytest.fixture()
+def with_extension(monkeypatch):
+    monkeypatch.setattr(os, "environ", {"LUMIGO_USE_TRACER_EXTENSION": "TRUE"})
```
saartochner, line 40: use `monkeypatch.setenv` (instead of override everything)

saartochner (approved):

PR #229, opened 2021-10-28
Rd 6294 python tracer extension with timeout flow is flaky

OrrLumigo:
The Extension now expects always two files:
XXX_span nd XXX_end
where XXX_span represents the start span as List[span] (length = 1) and  XXX_end that is List[span] (length = all the rest)
also, we removed the #DONE" from the suffix of files data since the extension validate the file stopped changing after T time

### DIFF
--- a/src/lumigo_tracer/lumigo_utils.py
+++ b/src/lumigo_tracer/lumigo_utils.py
@@ -94,7 +94,7 @@ def should_use_tracer_extension() -> bool:
 
 
 def get_extension_dir() -> str:
-    return (os.environ.get(LUMIGO_SPANS_DIR) or LUMIGO_SPANS_DIR).lower()
+    return (os.environ.get("LUMIGO_SPANS_DIR_KEY") or LUMIGO_SPANS_DIR).lower()
 
 
 def get_region() -> str:
@@ -396,7 +396,7 @@ def write_spans_to_files(
     spans: List[Dict], max_spans=MAX_NUMBER_OF_SPANS, is_start_span=True
 ) -> None:
     to_send = spans[:max_spans]
-    Path(LUMIGO_SPANS_DIR).mkdir(parents=True, exist_ok=True)
+    Path(get_extension_dir()).mkdir(parents=True, exist_ok=True)
     if is_start_span:
         get_logger().info("Creating start span file")
         write_extension_file(to_send, "span")
--- a/src/lumigo_tracer/lumigo_utils.py
+++ b/src/lumigo_tracer/lumigo_utils.py
@@ -94,7 +94,7 @@ def should_use_tracer_extension() -> bool:
 
 
 def get_extension_dir() -> str:
-    return (os.environ.get("LUMIGO_SPANS_DIR_KEY") or LUMIGO_SPANS_DIR).lower()
+    return (os.environ.get("LUMIGO_EXTENSION_SPANS_DIR_KEY") or LUMIGO_SPANS_DIR).lower()
 
 
 def get_region() -> str:
--- a/src/lumigo_tracer/lumigo_utils.py
+++ b/src/lumigo_tracer/lumigo_utils.py
@@ -75,7 +75,7 @@ NUMBER_OF_SPANS_IN_REPORT_OPTIMIZATION = 200
 DEFAULT_KEY_DEPTH = 4
 LUMIGO_TOKEN_KEY = "LUMIGO_TRACER_TOKEN"
 LUMIGO_USE_TRACER_EXTENSION = "LUMIGO_USE_TRACER_EXTENSION"
-EXTENSION_DIR = "/tmp/lumigo-spans"
+LUMIGO_SPANS_DIR = "/tmp/lumigo-spans"
 KILL_SWITCH = "LUMIGO_SWITCH_OFF"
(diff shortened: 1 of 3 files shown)


src/lumigo_tracer/lumigo_utils.py
```diff
@@ -377,26 +375,26 @@ def report_json(
     return duration
 
 
-def write_extension_file(data: dict, span_type: str):
-    to_send = aws_dump(data).encode() + EXTENSION_FILE_SUFFIX.encode()
+def write_extension_file(data: List[Dict], span_type: str):
+    to_send = aws_dump(data).encode()
     file_name = f"{hashlib.md5(to_send).hexdigest()}_{span_type}"
```
saartochner, line 380: why do you still have this md5?

src/test/unit/test_lumigo_utils.py
```diff
@@ -457,6 +456,13 @@ def test_get_edge_host(arg, host, monkeypatch):
 def test_report_json_extension_spans_mode(monkeypatch, reporter_mock):
     monkeypatch.setattr(Configuration, "should_report", True)
     monkeypatch.setenv("LUMIGO_USE_TRACER_EXTENSION", "TRUE")
+    mocked_urandom = MagicMock()
+    mocked_urandom.hex = MagicMock(return_value="my_mocked_data")
```
saartochner, line 460:
```suggestion
    mocked_urandom = MagicMock(hex=MagicMock(return_value="my_mocked_data"))
```

src/test/unit/test_lumigo_utils.py
```diff
@@ -457,6 +456,13 @@ def test_get_edge_host(arg, host, monkeypatch):
...
+    monkeypatch.setattr(os, "urandom", lambda *args, **kwargs: mocked_urandom)
+
+    start_span = [{"span": "true"}]
+    report_json(region=None, msgs=start_span, is_start_span=True)
```
saartochner, line 464: do you write files to the file system here?
This is highly not isolated. I.e. if you have these files and it stop working, then the next test will pass because the files are still there.
Please read: https://docs.pytest.org/en/6.2.x/tmpdir.html

src/test/unit/test_lumigo_utils.py
```diff
@@ -467,24 +473,16 @@ def test_report_json_extension_spans_mode(monkeypatch, reporter_mock):
...
-
-
-def get_span_file_name(span, _type):
-    to_send = aws_dump(span).encode() + EXTENSION_FILE_SUFFIX.encode()
+    start_file_path = get_span_file_name(start_span, "span")
+    end_file_path = get_span_file_name(spans, "end")
+    start_file_content = json.loads(open(start_file_path, "r").read())
+    end_file_content = json.loads(open(end_file_path, "r").read())
+    assert start_span == start_file_content
+    assert json.dumps(end_file_content) == json.dumps(spans)
```
saartochner, line 481: why do you compare strings here?
you can just assert `end_file_content == spans`, right?

saartochner (changes requested):
Great work!

src/test/unit/test_lumigo_utils.py
OrrLumigo: one is [{'0': 'a'}] and the other is [{0: 'a'}] so i dump it so its equal

src/lumigo_tracer/lumigo_utils.py
```diff
@@ -94,6 +93,10 @@ def should_use_tracer_extension() -> bool:
     return (os.environ.get(LUMIGO_USE_TRACER_EXTENSION) or "false").lower() == "true"
 
 
+def get_extension_dir() -> str:
+    return (os.environ.get(LUMIGO_SPANS_DIR) or LUMIGO_SPANS_DIR).lower()
```
saartochner, line 97:
```suggestion
    return (os.environ.get(LUMIGO_SPANS_DIR_KEY) or LUMIGO_SPANS_DIR).lower()
```

saartochner: and I would change it to something clear that this is for the extensions

src/test/unit/test_lumigo_utils.py
```diff
@@ -454,9 +453,18 @@ def test_get_edge_host(arg, host, monkeypatch):
     assert get_edge_host("region") == host
 
 
-def test_report_json_extension_spans_mode(monkeypatch, reporter_mock):
+def test_report_json_extension_spans_mode(monkeypatch, reporter_mock, tmpdir):
+    extension_dor = tmpdir.mkdir("tmp")
```
saartochner, line 457:
```suggestion
    extension_dir = tmpdir.mkdir("tmp")
```

src/test/unit/test_lumigo_utils.py
```diff
@@ -454,9 +453,18 @@ def test_get_edge_host(arg, host, monkeypatch):
...
+    monkeypatch.setattr(lumigo_utils, "get_extension_dir", lambda *args, **kwargs: extension_dor)
```
saartochner, line 458: more elegant way to do it is by `monkeypatch.setenv(LUMIGO_SPANS_DIR_KEY, extension_dir)`

saartochner (approved):
Small comments, please fix.
Approved in advanced

PR #235, opened 2021-12-13
feat: when using extension, Tracer shoudnt filter event data

OrrLumigo:


src/lumigo_tracer/event/event_dumper.py
```diff
@@ -251,7 +252,7 @@ def dump_event(
     ) -> str:
         max_size = Configuration.get_max_entry_size(has_error)
         if should_use_tracer_extension():
-            max_size = EVENT_MAX_SIZE
```
saartochner, line 254: can you add test here - what exactly do you want to happen/doesn't happen?

saartochner (approved):

PR #237, opened 2021-12-20
deprecate enhance print 

saartochner:


src/test/unit/test_tracer.py
```diff
@@ -183,28 +181,26 @@ def lambda_test_function(event, context):
...
+        assert any(line == "hello" for line in capturer.get_lines())
 
 
-def test_wrapping_without_print_override(context):
-    @lumigo_tracer()
+def test_wrapping_enhanced_print_backward_compatible(context):
+    @lumigo_tracer(enhance_print=True)
     def lambda_test_function(event, context):
         print("hello")
+        assert Configuration.enhanced_print is False
```
nirLumigo, line 199: I think that `Configuration` shouldn't have the `enhanced_print` property anymore, isn't it?

nirLumigo (changes requested):
Nice :)

src/test/unit/test_tracer.py
saartochner: I agree. Deleted :)

nirLumigo (approved):

saartochner:
Opened an identical PR under different name (ci purpose)

PR #239, opened 2022-01-24
add execution tags by env vars

saartochner:


src/lumigo_tracer/auto_tag/auto_tag_event.py
```diff
@@ -48,13 +48,25 @@ def auto_tag(event: dict):
                     add_execution_tag(key, headers[key])
 
 
+class ConfigurationHandler(EventAutoTagHandler):
+    @staticmethod
+    def is_supported(event) -> bool:
```
nirLumigo, line 53:
```suggestion
    def is_supported(event: dict) -> bool:
```
I think we should add enforcement for types also for this repo

src/lumigo_tracer/auto_tag/auto_tag_event.py
```diff
@@ -48,13 +48,25 @@ def auto_tag(event: dict):
...
+        return bool(event) and any(key in event for key in Configuration.auto_tag)
+
+    @staticmethod
+    def auto_tag(event: dict):
+        for key in Configuration.auto_tag:
+            if key in event:
+                add_execution_tag(key, event[key])
```
nirLumigo, line 60: Notice: I think that `execution_tag` support only strings, here the value might not be a string.

src/test/unit/test_tracer.py
```diff
@@ -341,6 +341,21 @@ def lambda_test_function(event, context):
...
+    @lumigo_tracer(auto_tag=[key])
+    def lambda_test_function(event, context):
+        return "ret_value"
+
+    result = lambda_test_function({key: value}, context)
+    assert result == "ret_value"
+    assert SpansContainer.get_span().function_span[EXECUTION_TAGS_KEY] == [
+        {"key": key, "value": value}
+    ]
+
```
nirLumigo, line 357: Add a test where the value isn't a string and make sure that it behaves as expected.

nirLumigo (changes requested):
Nice :)

src/lumigo_tracer/auto_tag/auto_tag_event.py
saartochner: currect, but `add_execution_tag` stringify the input. I'll add a test 👌

nirLumigo (approved):

PR #241, opened 2022-02-22
feat: add nested-auto-execution-tags

saartochner-lumigo:


src/lumigo_tracer/auto_tag/auto_tag_event.py
```diff
@@ -56,8 +56,12 @@ def is_supported(event: dict) -> bool:
     @staticmethod
     def auto_tag(event: dict):
         for key in Configuration.auto_tag:
-            if key in event:
-                add_execution_tag(key, event[key])
+            try:
+                value = safe_get(event, key.split("."))  # type: ignore
```
nirLumigo, line 60: The user won't be able to auto-tag keys that have "." in them. Are we ok with this?

nirLumigo (commented):
Nice! :)

src/lumigo_tracer/auto_tag/auto_tag_event.py
saartochner: yes!
we can support escaping, but this is future work IMO.
WDYT?

nirLumigo (approved):

PR #244, opened 2022-04-19
feat: add manual-trace feature

doriaviram:


README.md
```diff
@@ -143,7 +143,6 @@ Getting your machine ready to develop against the package is a straightforward p
 1. Create a virtual environment for the project `virtualenv venv -p python3`
 1. Activate the virtualenv: `. venv/bin/activate`
 1. Install dependencies: `pip install -r requirements.txt`
-1. Navigate to the source directory: `cd src` and
```
doriaviram, line 146: the setup is in the root dir

src/lumigo_tracer/spans_container.py
```diff
@@ -102,11 +103,13 @@ def __init__(
                 },
                 "isMalformedTransactionId": malformed_txid,
                 EXECUTION_TAGS_KEY: [],
+                "manualTraces": [],
```
saartochner, line 106:
```suggestion
                MANUAL_TRACES_KEY: [],
```

src/lumigo_tracer/spans_container.py
```diff
@@ -243,6 +246,18 @@ def get_tags_len(self) -> int:
...
         self.function_span[EXECUTION_TAGS_KEY].append({"key": key, "value": value})
 
+    def start_manual_trace(self, name: str) -> None:
+        now = int(datetime.now().timestamp() * 1000)
+        self.manual_trace_timers[name] = now
+
+    def stop_manual_trace(self, name: str) -> None:
+        manual_trace_started = self.manual_trace_timers.pop(name, None)
+        if manual_trace_started:
+            now = int(datetime.now().timestamp() * 1000)
```
saartochner, line 256: get_current_ms_time

src/lumigo_tracer/spans_container.py
```diff
@@ -102,11 +103,13 @@ def __init__(
                 },
                 "isMalformedTransactionId": malformed_txid,
                 EXECUTION_TAGS_KEY: [],
+                "manualTraces": [],
             },
             self.base_msg,
         )
         self.span_ids_to_send: Set[str] = set()
         self.spans: Dict[str, Dict] = {}
+        self.manual_trace_timers: Dict[str, int] = {}
```
saartochner, line 112: This name does not indicative enough. Maybe `manual_trace_start_times`?

src/test/unit/test_spans_container.py
```diff
@@ -204,6 +204,51 @@ def test_add_tag():
...
+def test_start_manual_trace_simple_flow():
+    SpansContainer.get_span().start_manual_trace("11")
+    SpansContainer.get_span().stop_manual_trace("11")
+
+    manual_tracers = SpansContainer.get_span().function_span[MANUAL_TRACES_KEY]
+    now = datetime.now().timestamp() * 1000
+    one_min_ago = (datetime.now() - timedelta(minutes=1)).timestamp() * 1000
+    assert len(manual_tracers) == 1
+    assert manual_tracers[0]["startTime"] >= one_min_ago
+    assert manual_tracers[0]["startTime"] <= now
```
saartochner, line 216:
```suggestion
    assert one_min_ago <= manual_tracers[0]["startTime"] <= now
```
same to below

src/test/unit/test_spans_container.py
```diff
@@ -204,6 +204,51 @@ def test_add_tag():
...
+    assert manual_tracers[0]["endTime"] >= one_min_ago
+    assert manual_tracers[0]["endTime"] <= now
```
saartochner, line 218:
```suggestion
    assert one_min_ago <= manual_tracers[0]["startTime"] < manual_tracers[0]["endTime"] <= now
```

src/test/unit/test_spans_container.py
```diff
@@ -204,6 +204,51 @@ def test_add_tag():
     ]
 
 
+def test_start_manual_trace_simple_flow():
+    SpansContainer.get_span().start_manual_trace("11")
+    SpansContainer.get_span().stop_manual_trace("11")
+
+    manual_tracers = SpansContainer.get_span().function_span[MANUAL_TRACES_KEY]
+    now = datetime.now().timestamp() * 1000
+    one_min_ago = (datetime.now() - timedelta(minutes=1)).timestamp() * 1000
```
saartochner, line 213: the interesting thing to check here is not "one minute ago", but just capture the timestamp in the beginning of the test. (just put `before= get_current_ms_time()` as the first line of the test)

src/lumigo_tracer/spans_container.py
```diff
@@ -243,6 +246,18 @@ def get_tags_len(self) -> int:
...
 
+    def start_manual_trace(self, name: str) -> None:
+        now = int(datetime.now().timestamp() * 1000)
+        self.manual_trace_timers[name] = now
+
+    def stop_manual_trace(self, name: str) -> None:
+        manual_trace_started = self.manual_trace_timers.pop(name, None)
+        if manual_trace_started:
+            now = int(datetime.now().timestamp() * 1000)
+            self.function_span[MANUAL_TRACES_KEY].append(
```
saartochner, line 257: I would actually put it in the `self.function_span[MANUAL_TRACES_KEY]` already when it starts.
It will help you to identify the spans that causes timeouts.
This function should only update the `endTime`

src/test/unit/test_spans_container.py
```diff
@@ -204,6 +204,51 @@ def test_add_tag():
...
+    SpansContainer.get_span().stop_manual_trace("11")
+    SpansContainer.get_span().stop_manual_trace("22")
+
+    manual_tracers = SpansContainer.get_span().function_span[MANUAL_TRACES_KEY]
+    now = datetime.now().timestamp() * 1000
+    one_min_ago = (datetime.now() - timedelta(minutes=1)).timestamp() * 1000
+    assert len(manual_tracers) == 2
+    for manual_trace in manual_tracers:
+        assert manual_trace["startTime"] >= one_min_ago
+        assert manual_trace["startTime"] <= now
```
saartochner, line 249: here again you can simplify and make the assertion much stronger IMO (e.g. check that the startTime of the 22 is lower than endTime of 11

src/lumigo_tracer/spans_container.py
saartochner: Actually it will not work because the function_span will not be ended in this case, which is a *real* bummer.
Why didn't you consider it as a span?

doriaviram: Its a good idea but its not supported yet, I will keep it as-is

saartochner (approved):

src/lumigo_tracer/user_utils.py
```diff
@@ -110,6 +110,25 @@ def validate_tag(key, value, tags_len, should_log_errors):
...
 
+def manual_trace(func):
+    def wrapper(*args, **kwargs):
+        with ManualTrace(func.__name__):
+            func(*args, **kwargs)
+
+    return wrapper
+
+
+class ManualTrace:
```
saartochner, line 121: you can write this code better with https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager

PR #247, opened 2022-05-01
feat: add flag to avoid collecting the HTTP's request and response body

saartochner:


efimk-lu:
@saartochner just making sure you are not developing things twice, what about `LUMIGO_VERBOSE`

saartochner:
> 

That's a great comment!
I went over the code, and there is a bit of a difference between the two: with `LUMIGO_VERBOSE=false`, we collect from the HTTP span only the host, method, messageId and resourceName. 
Here we also include headers and URI.
@uri-p you requested the feature, WDYT?

uri-p (approved):

PR #255, opened 2022-06-30
feat: support lambda -> sqs -> sns -> lambda

saartochner:


src/test/unit/event/test_event_trigger.py
```diff
@@ -139,6 +139,75 @@
...
+                "arn": "arn:aws:sqs:us-east-1:123456789:sqs-queue-name",
+                "messageIdToChainResource": [
+                    {
+                        "TopicArn": "arn:aws:sns:us-west-2:723663554526:tracer-test-saart-temp-Pttcj",
+                        "childMessageId": "f4ceb23d-2ae7-44d3-b171-df7ab2d10a81",
+                        "parentMessageId": "2c78f253-4cd9-57bb-8bc3-a965e40a293e",
+                        "resourceType": "sns",
+                    }
+                ],
+                "messageIds": [
```
nirLumigo, line 174: This is `"messageIds"` but in the next test it is `"messageId`, is this on purpose?

src/lumigo_tracer/event/event_trigger.py
```diff
@@ -256,6 +258,38 @@ def _parse_dynamomdb_event(event) -> Dict[str, Union[int, List[str]]]:
...
 
-def _parse_sqs_event(event) -> Dict[str, Union[int, List[str]]]:
-    mids = [record["messageId"] for record in event["Records"] if record.get("messageId")]
-    return {MESSAGE_IDS_KEY: mids} if len(mids) > 1 else {MESSAGE_ID_KEY: mids[0]}
+def _parse_sqs_event(event) -> Dict[str, Union[int, str, List[str], List[Dict[str, str]]]]:
+    message_ids = []
+    chained_resources: List[Dict[str, str]] = []
+    for record in event.get("Records", []):
+        record_message_id = None
+        if record.get("messageId"):
```
nirLumigo, line 266: IMO it will be nicer to do:
```
if record.get("messageId"):
    continue
```
and then you won't have to do `if record_message_id` later.

src/lumigo_tracer/event/event_trigger.py
```diff
@@ -256,6 +258,38 @@ def _parse_dynamomdb_event(event) -> Dict[str, Union[int, List[str]]]:
...
+            record_message_id = record["messageId"]
+            message_ids.append(record_message_id)
+        body = record.get("body", "")
+        if (
+            record_message_id
+            and isinstance(body, str)
+            and "SimpleNotificationService" in body
+            and "TopicArn" in body
+        ):
```
nirLumigo, line 275: Please extract it to a function (or add comment)  because it's hard to understand what happens here, maybe something like `_is_sns(body)`.

nirLumigo (changes requested):
Great! 💪

src/test/unit/event/test_event_trigger.py
saartochner: yes :)
That's how it was even before my change. This is for performance, but not very important

PR #258, opened 2022-08-24
Feature: send execution tags on timeout

saartochner:


src/lumigo_tracer/spans_container.py
```diff
@@ -137,6 +152,9 @@ def handle_timeout(self, *args):
             get_logger().info("The tracer reached the end of the timeout timer")
             spans_id_copy = self.span_ids_to_send.copy()
             to_send = [self.spans[span_id] for span_id in spans_id_copy]
+            enrichment_span = self.generate_enrichment_span()
+            if enrichment_span:
+                to_send.append(enrichment_span)
```
nirLumigo, line 157: Do you think that maybe it will be better to put it at the beginning of `to_send` instead of the end?
In case there are a lot of http calls and we don't send all of them, we still want to get the `enrichment_span`.

src/test/unit/test_spans_container.py
```diff
@@ -200,13 +201,35 @@ def test_timeout_mechanism_timeout_occurred_send_new_spans(monkeypatch, context,
     assert SpansContainer.get_span().span_ids_to_send
 
 
+def test_timeout_mechanism_timeout_occurred_but_finish_check_enrichment(
```
nirLumigo, line 204: 💪

src/test/unit/test_tracer.py
```diff
@@ -376,25 +376,25 @@ def lambda_test_function(event, context):
 
     result = lambda_test_function({}, context)
     assert result == "ret_value"
-    assert SpansContainer.get_span().function_span[EXECUTION_TAGS_KEY] == [
-        {"key": key, "value": value}
-    ]
+    final_send = reporter_mock.call_args_list[-1][1]["msgs"]
+    enrichment_span = next(s for s in final_send if s["type"] == ENRICHMENT_TYPE)
```
nirLumigo, line 380: Can you extract it to a testing-util function and use this function in all the relevant places?

nirLumigo (approved):
Great!

src/lumigo_tracer/spans_container.py
saartochner: Interesting! that's a great idea!!

PR #259, opened 2022-08-28
Rd 9037 increase robustness

saartochner:
Two features:
1. lumigo_safe_execute now gets severity
2. identify non-aws services in the httpParser (reduce false positive)

### DIFF
--- a/src/lumigo_tracer/lumigo_utils.py
+++ b/src/lumigo_tracer/lumigo_utils.py
@@ -386,7 +386,7 @@ def report_json(
         internal_analytics_message("report: socket.timeout")
     except Exception as e:
         if should_retry:
-            get_logger().info(f"Could not report to {host}. Retrying.", exc_info=e)
+            get_logger().info(f"Could not report to {host}: ({str(e)}). Retrying.")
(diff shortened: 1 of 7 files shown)


src/lumigo_tracer/lumigo_utils.py
```diff
@@ -386,7 +386,7 @@ def report_json(
         internal_analytics_message("report: socket.timeout")
     except Exception as e:
         if should_retry:
-            get_logger().info(f"Could not report to {host}. Retrying.", exc_info=e)
+            get_logger().info(f"Could not report to {host}: ({str(e)}). Retrying.")
```
moshe-shaham-lumigo, line 389: why is this change? isn't better to let the info function handle the exception display?

moshe-shaham-lumigo (approved):

src/lumigo_tracer/lumigo_utils.py
saartochner: This is a legitimate flow, and the exception is really terrifying. 
Generally you're right that we better have more data, but here I prefer to not bomb the logs :)

PR #260, opened 2022-08-31
feat: add w3c headers to http request

saartochner:


src/lumigo_tracer/wrappers/http/http_parser.py
```diff
@@ -61,14 +62,18 @@ def parse_request(self, parse_params: HttpRequest) -> dict:
                 "instance_id": parse_params.instance_id,
             }
 
+        message_id = None
+        if parse_params.headers and is_w3c_headers(parse_params.headers):
+            message_id = get_w3c_message_id(parse_params.headers)
```
nirLumigo, line 67: Where did the message-id come from? Don't we usually get message-id in the response?

src/lumigo_tracer/wrappers/http/sync_http_wrappers.py
```diff
@@ -245,10 +246,15 @@ def _headers_reminder_wrapper(func, instance, args, kwargs):
     Remember the headers helps us to improve performances on requests that use this flow.
     """
     with lumigo_safe_execute("add hooked data"):
+        headers = safe_get_list(args, HEADERS_ARG_INDEX_REQUEST) or kwargs.get("headers") or {}
+        if Configuration.propagate_w3c:
+            SpansContainer.get_span().add_w3c_trace_propagator(headers)
+            kwargs["headers"] = headers
+            args = args[:HEADERS_ARG_INDEX_REQUEST]
```
nirLumigo, line 253: Why removing the args after the headers?

src/lumigo_tracer/wrappers/http/sync_http_wrappers.py
```diff
@@ -245,10 +246,15 @@ def _headers_reminder_wrapper(func, instance, args, kwargs):
     Remember the headers helps us to improve performances on requests that use this flow.
     """
     with lumigo_safe_execute("add hooked data"):
+        headers = safe_get_list(args, HEADERS_ARG_INDEX_REQUEST) or kwargs.get("headers") or {}
+        if Configuration.propagate_w3c:
+            SpansContainer.get_span().add_w3c_trace_propagator(headers)
+            kwargs["headers"] = headers
```
nirLumigo, line 252: The function `add_w3c_trace_propagator` already modifies the `headers` object so I think that this line is redundant.

src/test/unit/test_lumigo_utils.py
```diff
@@ -400,6 +400,13 @@ def test_config_step_function_without_envs(monkeypatch, configuration_value):
     assert Configuration.is_step_function == configuration_value
 
 
+@pytest.mark.parametrize("value, expected", (("TRUE", True), ("FALSE", False)))
+def test_config_propagate_w3c_by_env(monkeypatch, value, expected):
+    monkeypatch.setenv("LUMIGO_PROPAGATE_W3C", value)
```
nirLumigo, line 405: Can you add another test for the default value? Where the env var "LUMIGO_PROPAGATE_W3C" isn't set.

src/test/unit/test_tracer.py
```diff
@@ -228,6 +231,51 @@ def lambda_test_function(event, context):
...
+    def lambda_test_function(event, context):
+        conn = http.client.HTTPConnection("www.google.com")
+        conn.request("POST", "/", json.dumps({"a": "b"}))
+        return {"hello": "world"}
+
+    lambda_test_function({}, context)
+    http_spans = list(SpansContainer.get_span().spans.values())
+    assert (
+        TRACEPARENT_HEADER_NAME in http_spans[0]["info"]["httpInfo"]["request"]["headers"]
+    ) == propagate_w3c
```
nirLumigo, line 246: To this tests and all the other related tests, can you also verify:
1. The value of `http_spans[0]["info"]["httpInfo"]["request"]["headers"]`.
2. That `conn.request` was called with all the correct argument.

src/test/unit/test_tracer.py
```diff
@@ -228,6 +231,51 @@ def lambda_test_function(event, context):
...
+
+
+def test_add_w3c_headers_to_http_with_headers_as_args(monkeypatch, context, aws_env):
+    monkeypatch.setenv(LUMIGO_PROPAGATE_W3C, "TRUE")
+
+    @lumigo_tracer()
+    def lambda_test_function(event, context):
+        conn = http.client.HTTPConnection("www.google.com")
+        conn.request("POST", "/", json.dumps({"a": "b"}), {"another": "header"})
```
nirLumigo, line 255: Can you add an argument after the headers?

src/test/unit/test_w3c_context.py
```diff
@@ -0,0 +1,80 @@
+import pytest
```
nirLumigo, line 1: If I understood correctly, the code was copied from OTL, was the test file also copied? If so, please add a comment.

src/lumigo_tracer/w3c_context.py
```diff
@@ -0,0 +1,59 @@
+import re
+import random
+from typing import Dict, Optional
+
+# parts of this files were copied from:
```
nirLumigo, line 5: Shouldn't it be in the "libs" directory?

src/test/unit/wrappers/http/test_http_parser.py
```diff
@@ -353,3 +354,36 @@ def test_sns_parser_resource_name_target_arn():
...
+        headers={
+            TRACEPARENT_HEADER_NAME: "00-11111111111111111111111100000000-aaaaaaaaaaaaaaaa-01"
+        },
+        body=b"TargetArn=arn:aws:sns:us-west-2:123456:sns-name",
+    )
+    response = parser.parse_request(params)
+    assert response["info"]["messageId"] == "aaaaaaaaaaaaaaaa"
+
+
+def test_parser_w3c_weaker_then_other_message_id():
```
nirLumigo, line 374: Can you add a docstring explaining it? What is "weaker" in this context?

src/test/unit/wrappers/http/test_http_parser.py
```diff
@@ -353,3 +354,36 @@ def test_sns_parser_resource_name_target_arn():
...
+        headers={
+            "x-amz-target": "DynamoDB_20120810.PutItem",
+            TRACEPARENT_HEADER_NAME: "00-11111111111111111111111100000000-aaaaaaaaaaaaaaaa-01",
+        },
+        body=json.dumps({"TableName": "resourceName", "Item": {"key": {"S": "value"}}}),
+    )
+    response = parser.parse_request(params)
+    assert response["info"]["resourceName"] == "resourceName"
+    assert response["info"]["dynamodbMethod"] == "PutItem"
+    assert response["info"]["messageId"] == "1ad3dccc8064a706957c2c06ce3796bb"
```
nirLumigo, line 389: Where did this id come from?

nirLumigo (changes requested):
Nice! :)

src/lumigo_tracer/wrappers/http/http_parser.py
saartochner: That's the beauty of OTEL - the parent tells the child what his txid and parentSpanId is :-)

src/lumigo_tracer/wrappers/http/sync_http_wrappers.py
saartochner: This is deep!
This is the signature of the wrapped function: 
image
We want to handle both cases: whether the headers was given as args or as kwargs. In the following lines, we use the function *differently*, and move the headers as kwargs!
Therefore, if the users has passed `headers` to `HTTPConnection.request` with args - we trim it. If he passed it as kwarg, we do here nothing, and just replacing the value in the following line

saartochner: This is deep!
This is the signature of the wrapped function: 
image
We want to handle both cases: whether the headers was given as args or as kwargs. In the following lines, we use the function *differently*, and move the headers as kwargs!
Therefore, if the users has passed `headers` to `HTTPConnection.request` with args - we trim it. If he passed it as kwarg, we do here nothing, and just replacing the value in the following line

src/test/unit/wrappers/http/test_http_parser.py
saartochner: Ohh this was just copied from the ddb test - this is hash on keys of the ddb item

src/lumigo_tracer/w3c_context.py
saartochner: This is a misleading comment, only a small part was copied. I'll rephrase it!

src/test/unit/test_tracer.py
saartochner: Take a look! This is an interesting test 🤔

src/lumigo_tracer/wrappers/http/sync_http_wrappers.py
saartochner: Nope! Look at the previous comment :)
(This is the case where no headers were given!

nirLumigo (approved):

PR #263, opened 2022-10-04
avoid return-value warning when the runtime version is not python3.7

saartochner:
The logic of how to serialize the lambda's return value has been changes from between python3.7 and python3.8.
In the tracer, if we can't serialize the result, we will attach an error to the function_span to notify the user. 
Therefore, we should match our logic to the runtime. This PR is the first to do so.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -19,6 +19,7 @@ repos:
     hooks:
     -   id: flake8
         args: ["--ignore","E501,W503","--exclude","src/lumigo_tracer/libs/*"]
+        additional_dependencies: ['importlib-metadata==4.13.0']
 
 -   repo: https://github.com/PyCQA/bandit
     rev: '1.7.0'
@@ -26,6 +27,7 @@ repos:
     -   id: bandit
         exclude: ^src/test/
         args: ["-lll"]
+        additional_dependencies: ['importlib-metadata==4.13.0']
 
 -   repo: https://github.com/Yelp/detect-secrets
     rev: v1.2.0
--- a/src/test/unit/test_tracer.py
+++ b/src/test/unit/test_tracer.py
@@ -1,3 +1,5 @@
+import logging
+
 import datetime
 import json
 import re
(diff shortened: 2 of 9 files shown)


.pre-commit-config.yaml
```diff
@@ -19,13 +19,15 @@ repos:
     hooks:
     -   id: flake8
         args: ["--ignore","E501,W503","--exclude","src/lumigo_tracer/libs/*"]
+        additional_dependencies: ['importlib-metadata==4.13.0']
```
saartochner, line 22: fixing the cicd error `AttributeError: 'EntryPoints' object has no attribute 'items'`

.pre-commit-config.yaml
```diff
@@ -19,13 +19,15 @@ repos:
...
 
 -   repo: https://github.com/PyCQA/bandit
     rev: '1.7.0'
     hooks:
     -   id: bandit
         exclude: ^src/test/
         args: ["-lll"]
+        additional_dependencies: ['importlib-metadata==4.13.0']
```
saartochner, line 30: fixing the cicd error `AttributeError: 'EntryPoints' object has no attribute 'items'`

src/test/unit/test_tracer.py
```diff
@@ -468,6 +472,23 @@ def lambda_test_function(event, context):
...
+        return {"a": datetime.datetime.now()}
+
+    lambda_test_function({}, context)
+
+    function_span = SpansContainer.get_span().function_span
+    assert function_span["return_value"] is None
+    assert "error" not in function_span
+    log = caplog.records[-1]
+    assert log.levelno == logging.ERROR
+    assert "Could not serialize the return value of the lambda" in log.message
```
nirLumigo, line 489: IMO it's better to do something like this because then we don't care about the order of the logs.
```suggestion
    next(log for log in caplog.records if log.levelno == logging.ERROR and "Could not serialize the return value of the lambda" in log.message)
```

nirLumigo (approved):
Nice! :)

PR #271, opened 2023-02-19
python-core: migrate trigger utils

saartochner:
all the trigger prasers and their dependencies moved to python-core.
Including the logger, some parsing and time utils, configuration, and many tests


setup.py
```diff
@@ -14,7 +14,7 @@
     url="https://github.com/lumigo-io/python_tracer.git",
     package_dir={"": "src"},
     packages=setuptools.find_packages("src"),
-    install_requires=[],
+    install_requires=["lumigo_core==0.0.4"],
```
saartochner, line 17: TODO: Create a stable version and use it here

src/test/unit/lambda_tracer/test_tracer.py
```diff
@@ -186,30 +185,28 @@ def lambda_test_function(event, context):
...
-    with CaptureOutput() as capturer:
-        assert lambda_test_function({}, context) == 1
-        assert any(line == "hello" for line in capturer.get_lines())
+    assert lambda_test_function({}, context) == 1
+    assert "hello" in capsys.readouterr().out
 
 
 @pytest.mark.parametrize("is_verbose", [True, False])
-def test_skip_collecting_http_parts(monkeypatch, context, is_verbose):
+def test_skip_collecting_http_parts(wrap_all_libraries, monkeypatch, context, is_verbose):
```
saartochner, line 209: This is an unrelated change - I noticed that this test doesn't pass if you run it by itself (it passed due to a dependency leak from another test). I just made sure that it'll have all its dependencies.

src/test/unit/lambda_tracer/test_tracer.py
```diff
@@ -186,30 +185,28 @@ def lambda_test_function(event, context):
...
+def test_wrapping_enhanced_print_backward_compatible(context, capsys):
     @lumigo_tracer(enhance_print=True)
     def lambda_test_function(event, context):
         print("hello")
         return 1
 
-    with CaptureOutput() as capturer:
-        assert lambda_test_function({}, context) == 1
-        assert any(line == "hello" for line in capturer.get_lines())
+    assert lambda_test_function({}, context) == 1
```
saartochner, line 204: unrelated change as well - I removed this external dependency and used a native solution. It is also faster

src/lumigo_tracer/event/trigger_parsing/__init__.py
```diff
@@ -1,37 +1,5 @@
-import re
-from typing import List, Type
+from lumigo_tracer.event.trigger_parsing.step_function_parser import StepFunctionEventTriggerParser
```
saartochner, line 1: All trigger parsers moved to the core, besides the step function which doesn't belong there

GuyMoses (approved):

PR #285, opened 2023-08-13
Rd 11126 lambda python node tracers sqs json support  python only

sagivoululumigo:
# Feature
* Support for parsing SQS messages sent in the new JSON protocol (see [docs](https://aws.amazon.com/about-aws/whats-new/2023/07/json-protocol-support-amazon-sqs/))

### DIFF
--- a/src/lumigo_tracer/wrappers/http/http_parser.py
+++ b/src/lumigo_tracer/wrappers/http/http_parser.py
@@ -447,7 +447,7 @@ def get_parser(host: str, headers: Optional[dict] = None) -> Type[Parser]:  # ty
     # SQS Legacy Endpoints: https://docs.aws.amazon.com/general/latest/gr/rande.html
     elif service in ("sqs", "sqs-fips") or "queue.amazonaws.com" in host:
         using_json_protocol = (
-            _headers.get("content-type", "").lower() == "application/x-amz-json-1.0"
+            _headers.get("content-type", "").lower().startswith("application/x-amz-json-")
         )
         return SqsJsonParser if using_json_protocol else SqsXmlParser
     elif "execute-api" in host:
--- a/src/lumigo_tracer/wrappers/http/http_parser.py
+++ b/src/lumigo_tracer/wrappers/http/http_parser.py
@@ -1,6 +1,6 @@
-from abc import ABC, abstractmethod
 import json
 import uuid
+from abc import ABC, abstractmethod
 from typing import List, Optional, Type
 from urllib.parse import parse_qsl, unquote, urlencode, urlparse, urlunparse
 
@@ -112,7 +112,7 @@ class Parser:
     def scrub_query_params(uri: str) -> str:
         with lumigo_safe_execute("scrub_query_params"):
             regexes = (
-                CoreConfiguration.secret_masking_regex_http_query_params or get_omitting_regex()
+                    CoreConfiguration.secret_masking_regex_http_query_params or get_omitting_regex()
             )
             if not uri or "?" not in uri or not regexes:
                 return uri
@@ -132,7 +132,8 @@ class ServerlessAWSParser(Parser):
     # Override this field to add message id using the amz headers
     should_add_message_id = True
 
-    def parse_response(self, url: str, status_code: int, headers, body: bytes) -> dict:  # type: ignore[no-untyped-def,type-arg]
+    def parse_response(self, url: str, status_code: int, headers,
+                       body: bytes) -> dict:  # type: ignore[no-untyped-def,type-arg]
         additional_info = {}
         message_id = headers.get("x-amzn-requestid")
         if message_id and self.should_add_message_id:
@@ -210,7 +211,8 @@ class SnsParser(ServerlessAWSParser):
             super().parse_request(parse_params),
         )
 
-    def parse_response(self, url: str, status_code: int, headers, body: bytes) -> dict:  # type: ignore[no-untyped-def,type-arg]
+    def parse_response(self, url: str, status_code: int, headers,
+                       body: bytes) -> dict:  # type: ignore[no-untyped-def,type-arg]
         return recursive_json_join(  # type: ignore[no-any-return]
             {
                 "info": {
@@ -244,7 +246,8 @@ class KinesisParser(ServerlessAWSParser):
             super().parse_request(parse_params),
         )
 
-    def parse_response(self, url: str, status_code: int, headers, body: bytes) -> dict:  # type: ignore[no-untyped-def,type-arg]
+    def parse_response(self, url: str, status_code: int, headers,
+                       body: bytes) -> dict:  # type: ignore[no-untyped-def,type-arg]
         return recursive_json_join(  # type: ignore[no-any-return]
             {"info": {"messageId": KinesisParser._extract_message_id(body)}},
             super().parse_response(url, status_code, headers, body),
@@ -269,7 +272,8 @@ class BaseSqsParser(ServerlessAWSParser, ABC):
             super().parse_request(parse_params),
         )
 
-    def parse_response(self, url: str, status_code: int, headers, body: bytes) -> dict:  # type: ignore[no-untyped-def,type-arg]
+    def parse_response(self, url: str, status_code: int, headers,
+                       body: bytes) -> dict:  # type: ignore[no-untyped-def,type-arg]
         return recursive_json_join(  # type: ignore[no-any-return]
             {"info": {"messageId": self._extract_message_id(body)}},
             super().parse_response(url, status_code, headers, body),
@@ -351,7 +355,8 @@ class S3Parser(Parser):
             super().parse_request(parse_params),
         )
 
-    def parse_response(self, url: str, status_code: int, headers, body: bytes) -> dict:  # type: ignore[no-untyped-def,type-arg]
+    def parse_response(self, url: str, status_code: int, headers,
+                       body: bytes) -> dict:  # type: ignore[no-untyped-def,type-arg]
         return recursive_json_join(  # type: ignore[no-any-return]
             {"info": {"messageId": headers.get("x-amz-request-id")}},
             super().parse_response(url, status_code, headers, body),
@@ -377,7 +382,8 @@ class EventBridgeParser(Parser):
             super().parse_request(parse_params),
         )
 
-    def parse_response(self, url: str, status_code: int, headers, body: bytes) -> dict:  # type: ignore[no-untyped-def,type-arg]
+    def parse_response(self, url: str, status_code: int, headers,
+                       body: bytes) -> dict:  # type: ignore[no-untyped-def,type-arg]
         try:
             parsed_body = json.loads(body)
         except json.JSONDecodeError as e:
@@ -395,7 +401,8 @@ class EventBridgeParser(Parser):
 class ApiGatewayV2Parser(ServerlessAWSParser):
     # API-GW V1 covered by ServerlessAWSParser
 
-    def parse_response(self, url: str, status_code: int, headers, body: bytes) -> dict:  # type: ignore[no-untyped-def,type-arg]
+    def parse_response(self, url: str, status_code: int, headers,
+                       body: bytes) -> dict:  # type: ignore[no-untyped-def,type-arg]
         aws_request_id = headers.get("x-amzn-requestid")
         apigw_request_id = headers.get("apigw-requestid")
         message_id = aws_request_id or apigw_request_id
--- a/src/lumigo_tracer/wrappers/http/http_parser.py
+++ b/src/lumigo_tracer/wrappers/http/http_parser.py
@@ -1,3 +1,4 @@
+from abc import ABC, abstractmethod
 import json
 import uuid
 from typing import List, Optional, Type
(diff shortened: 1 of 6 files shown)


src/lumigo_tracer/wrappers/http/http_parser.py
```diff
@@ -289,6 +308,45 @@ def _extract_message_id(response_body: bytes) -> Optional[str]:
...
         )
 
+    @staticmethod
+    def _extract_queue_url(request_body: bytes) -> Optional[str]:
+        return safe_key_from_query(request_body, "QueueUrl")
+
+
+class SqsJsonParser(BaseSqsParser):
+    @staticmethod
+    def _extract_message_id(response_body: bytes) -> Optional[str]:
```
GuyMoses, line 318: is `response_body` really bytes?
you're trying to `json.loads` it

src/test/unit/wrappers/http/test_http_parser.py
```diff
@@ -122,18 +127,134 @@ def test_lambda_parser_resource_name(uri, resource_name):
...
+            # Send single message to SQS response
+            b"{"
+            b'   "MD5OfMessageAttributes":"6e6aba56e93b3ddfdfe3fa28895feece",'  # pragma: allowlist secret
+            b'   "MD5OfMessageBody":"0d40eb1479f7e61b1a1c7a425c3949e4",'  # pragma: allowlist secret
+            b'   "MessageId":"c5aca29a-ff2f-4db5-94c3-90523d1ed4ca"'  # pragma: allowlist secret
+            b"}",
+            "c5aca29a-ff2f-4db5-94c3-90523d1ed4ca",
+        ),
+        (
+            # Send batch message request to SQS (on record in the batch), successful message
```
GuyMoses, line 183: add a a test for more than 1 message in the batch

GuyMoses (changes requested):

src/lumigo_tracer/wrappers/http/http_parser.py
```diff
@@ -289,6 +308,45 @@ def _extract_message_id(response_body: bytes) -> Optional[str]:
...
+        # This should work if the request was to send a batch of messages.
+        # Find the message id of the first successful message, and if there are none the first failed message
+        messages = parsed_body.get("Successful", []) + parsed_body.get("Failed", [])
+        if (
+            messages
+            and isinstance(messages[0], dict)
+            and messages[0].get("MessageId")
+            and isinstance(messages[0].get("MessageId"), str)
+        ):
+            return messages[0].get("MessageId")
```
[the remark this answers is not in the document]
sagivoululumigo: No... I asked and (I think Saar) told me that we know and it is good enough for now (This is how it's implemented in XML as well in python & node tracer)

src/test/unit/wrappers/http/test_http_parser.py
sagivoululumigo: Done

GuyMoses (approved):

src/lumigo_tracer/wrappers/http/http_parser.py
```diff
@@ -260,19 +265,33 @@ def get_omit_skip_path() -> Optional[List[str]]:
         return ["PartitionKey"]
 
 
-class SqsParser(ServerlessAWSParser):
+class BaseSqsParser(ServerlessAWSParser, ABC):
```
saartochner, line 268: Why did you use inheritance here? I don't see any common code that worths this extra complexity.
Please KISS the tracer (https://he.wikipedia.org/wiki/KISS)

src/lumigo_tracer/wrappers/http/http_parser.py
```diff
@@ -289,6 +308,45 @@ def _extract_message_id(response_body: bytes) -> Optional[str]:
...
+            get_logger().warning("Error while trying to parse sqs json request body", exc_info=e)
+
+        # If the request was to send a single message this should work
+        message_id = parsed_body.get("MessageId")
+        if message_id and isinstance(message_id, str):
+            return message_id
+
+        # This should work if the request was to send a batch of messages.
+        # Find the message id of the first successful message, and if there are none the first failed message
+        messages = parsed_body.get("Successful", []) + parsed_body.get("Failed", [])
```
saartochner, line 332: If a message has failed to be sent, why do we need its messageId? 🤔

src/lumigo_tracer/wrappers/http/http_parser.py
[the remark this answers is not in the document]
saartochner: Agree. Migrating to `MessageIds` might cause extra BE work. 
Si it depends on your task' goals :)

src/lumigo_tracer/wrappers/http/http_parser.py
```diff
@@ -354,12 +418,18 @@ def parse_response(self, url: str, status_code: int, headers, body: bytes) -> di
         )
 
 
-def get_parser(url: str, headers: Optional[dict] = None) -> Type[Parser]:  # type: ignore[type-arg]
+def get_parser(host: str, headers: Optional[dict] = None) -> Type[Parser]:  # type: ignore[type-arg]
+    # Headers are case-insensitive, so lets lowercase them and always search them in lowercase
+    lowercase_headers = {}
```
saartochner, line 423: probably dup with 
https://github.com/lumigo-io/python_tracer/blob/a48db8acfd1b353d0474809597c6be9f271ead96/src/lumigo_tracer/wrappers/http/sync_http_wrappers.py#L116

src/lumigo_tracer/wrappers/http/http_parser.py
```diff
@@ -370,11 +440,14 @@ def get_parser(url: str, headers: Optional[dict] = None) -> Type[Parser]:  # typ
...
-    elif safe_split_get(url, ".", 1) == "s3" or safe_split_get(url, ".", 0) == "s3":
+    elif safe_split_get(host, ".", 1) == "s3" or safe_split_get(host, ".", 0) == "s3":
         return S3Parser
     # SQS Legacy Endpoints: https://docs.aws.amazon.com/general/latest/gr/rande.html
-    elif service in ("sqs", "sqs-fips") or "queue.amazonaws.com" in url:
-        return SqsParser
-    elif "execute-api" in url:
+    elif service in ("sqs", "sqs-fips") or "queue.amazonaws.com" in host:
+        using_json_protocol = (
+            lowercase_headers.get("content-type", "").lower() == "application/x-amz-json-1.0"
```
saartochner, line 448: `1.0` is the version?
Do we want that this `if` will be false when they upgrade version?
I think that `SqsJsonParser` is a better fallback to `"application/x-amz-json-1.1"` than `SqsXmlParser`

src/test/component/test_component.py
```diff
@@ -23,6 +25,36 @@ def serverless_yaml():
     )
 
 
+@pytest.fixture()
```
saartochner, line 28: These are component tests.
I do want that they will interact with the internet.
Is there any particular reason for this change?

src/test/unit/wrappers/http/test_http_parser.py
```diff
@@ -122,18 +127,191 @@ def test_lambda_parser_resource_name(uri, resource_name):
...
+    "response_body, message_id",
+    [
+        (
+            # Send single message to SQS response
+            b"{"
+            b'   "MD5OfMessageAttributes":"6e6aba56e93b3ddfdfe3fa28895feece",'  # pragma: allowlist secret
+            b'   "MD5OfMessageBody":"0d40eb1479f7e61b1a1c7a425c3949e4",'  # pragma: allowlist secret
+            b'   "MessageId":"c5aca29a-ff2f-4db5-94c3-90523d1ed4ca"'  # pragma: allowlist secret
+            b"}",
+            "c5aca29a-ff2f-4db5-94c3-90523d1ed4ca",
```
saartochner, line 180: don't you prefer to use dict and `json.dumps` it? These binary strings looks very uncomfortable haha

src/test/unit/wrappers/http/test_http_parser.py
```diff
@@ -122,18 +127,191 @@ def test_lambda_parser_resource_name(uri, resource_name):
...
 
+@pytest.mark.parametrize(
+    "response_body, message_id",
+    [
+        (
+            # Send single message to SQS response
+            b"{"
+            b'   "MD5OfMessageAttributes":"6e6aba56e93b3ddfdfe3fa28895feece",'  # pragma: allowlist secret
+            b'   "MD5OfMessageBody":"0d40eb1479f7e61b1a1c7a425c3949e4",'  # pragma: allowlist secret
+            b'   "MessageId":"c5aca29a-ff2f-4db5-94c3-90523d1ed4ca"'  # pragma: allowlist secret
```
saartochner, line 178: usually I prefer to avoid random strings for readability. You can change it to any arbitrary value 🤷 
Then, it will be easier to understand that line 180 equals exactly line 178.
I would suggest using "aaaaa" and "bbbb"

src/test/unit/wrappers/http/test_http_parser.py
```diff
@@ -122,18 +127,191 @@ def test_lambda_parser_resource_name(uri, resource_name):
...
+            b'   "Successful":['
+            b"     {"
+            b'        "Id":"1",'
+            b'        "MD5OfMessageBody":"68390233272823b7adf13a1db79b2cd7",'  # pragma: allowlist secret
+            b'        "MessageId":"c5aca29a-ff2f-4db5-94c3-90523d1ed4ca"'
+            b"     },"
+            b"     {"
+            b'        "Id":"2",'
+            b'        "MD5OfMessageBody":"68390233272823b7adf13a1db7222222",'  # pragma: allowlist secret
+            b'        "MessageId":"c5aca29a-ff2f-4db5-94c3-90523d222222"'
```
saartochner, line 210: please change the string completely. this is super not readable that you change only the last 6 digits

src/test/unit/wrappers/http/test_http_parser.py
```diff
@@ -122,18 +127,191 @@ def test_lambda_parser_resource_name(uri, resource_name):
...
+            b'        "MD5OfMessageBody":"68390233272823b7adf13a1db7333333",'  # pragma: allowlist secret
+            b'        "MessageId":"c5aca29a-ff2f-4db5-94c3-90523d333333"'
+            b"     },"
+            b"     {"
+            b'        "Id":"4",'
+            b'        "MD5OfMessageBody":"68390233272823b7adf13a1db7444444",'  # pragma: allowlist secret
+            b'        "MessageId":"c5aca29a-ff2f-4db5-94c3-90523d444444"'
+            b"     }"
+            b"]}",
+            "c5aca29a-ff2f-4db5-94c3-90523d1ed4ca",
```
saartochner, line 263: I think that you missed the case that you have only failed

src/lumigo_tracer/wrappers/http/http_parser.py
sagivoululumigo: The `ServerlessAWSParser` inheritence is required because the `parse_request` & `parse_response` functions call `super().parse_request/parse_response` (the sqs parser adds info about the sqs queue, and the aws parser parses many other properties about the request like amazon request id etc...)

The ABC ingeritence is required because every sqs parser needs to have `_extract_message_id` & `_extract_queue_url` funcs to add information to the request & response in the `parse_request` & `parse_response`, but the implementation for these funcs is different in xml & json sqs parsing.

I can't use the `_extract_message_id` & `_extract_queue_url` functions in the `BaseSqsParser` class without making the class abstract & defining the abstract methods that the children will implement

src/lumigo_tracer/wrappers/http/http_parser.py
sagivoululumigo: This is the current implementation for SQS xml parsing so I copied it.

BTW it doesn't hurt to add the (true I might add) piece of information to the parsing of the request, it could be used in a further part of the pipeline if required

src/lumigo_tracer/wrappers/http/http_parser.py
sagivoululumigo: Im parsing specifically for `1.0` version because that is what the AWS documentation specifies must be in the HTTP request for SQS json request (see https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-making-api-requests-json.html#structure-post-request)

And about the default parsing, currently XML is the default protocol for communicating with SQS, and amazon wrote in their blog post that only once they see a large enough amount of users move to json that they will make it the default in their client libraries. So if most cases will use XML parsing, I want xml to be the default option (In the future if json will become the default we can make json the default)

src/test/component/test_component.py
sagivoululumigo: Yes, httpbin is flaky and returns gateway timeout errors. This solution creates an HTTP server that runs on another port on localhost, that any http library can communicate with

saartochner (changes requested):
Great work!

src/lumigo_tracer/wrappers/http/http_parser.py
saartochner: I meant that you should change line 448 to:
```
            lowercase_headers.get("content-type", "").lower().startswith("application/x-amz-json-")
```

src/lumigo_tracer/wrappers/http/http_parser.py
saartochner: The inheritance itself is useless.
You can do the same work with fewer lines of code and less architectural complexity.
Just remove the Base class

src/lumigo_tracer/wrappers/http/http_parser.py
saartochner: 1. the current implementation does not have failed messages 🤔 
2. adding unused information is wasteful

src/test/component/test_component.py
```diff
@@ -23,6 +25,36 @@ def serverless_yaml():
...
+def echo_server(httpserver: HTTPServer) -> HTTPServer:
+    def respond_like_httpbin_anything_path(request: Request) -> Response:
+        response_body = {
+            "data": request.data.decode("utf-8")
+            if isinstance(request.data, bytes)
+            else request.data,
+            "headers": dict(request.headers),
+        }
+        print(f"Echo server will return response body: {response_body}")
```
saartochner, line 37: remove prints

src/test/component/test_component.py
saartochner: ok then 👍

src/test/component/test_component.py
```diff
@@ -6,6 +6,8 @@
 import boto3
 import pytest
 from lumigo_core.lumigo_utils import md5hash
+from pytest_httpserver.httpserver import HTTPServer
+from werkzeug.wrappers import Request, Response
```
saartochner, line 10: reminder - add to requirements

src/lumigo_tracer/wrappers/http/http_parser.py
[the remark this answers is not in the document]
sagivoululumigo: Yes this isn't a part of my tasks goal

src/lumigo_tracer/wrappers/http/http_parser.py
sagivoululumigo: My bad, I thought the current implementation did do this.
Removed the failed message id

src/lumigo_tracer/wrappers/http/http_parser.py
sagivoululumigo: Yes I saw this, but because the lowercase keys aren't promised in the function params I worry that the function can be called from somewhere else without lowercasing the keys...

What i will do is to  add a note to the functions docstring to make sure other people know lowercase the header keys.

src/test/unit/wrappers/http/test_http_parser.py
sagivoululumigo: Done

src/test/unit/wrappers/http/test_http_parser.py
sagivoululumigo: Done

src/test/unit/wrappers/http/test_http_parser.py
sagivoululumigo: Done

src/test/component/test_component.py
sagivoululumigo: Done

src/test/unit/wrappers/http/test_http_parser.py
sagivoululumigo: I did check for only failed...

src/test/component/test_component.py
sagivoululumigo: Done

src/lumigo_tracer/wrappers/http/http_parser.py
sagivoululumigo: Done

saartochner (approved):

PR #299, opened 2023-12-25
feat: add support for python 3.12 lambdas

sagivoululumigo:


saartochner (approved):

sagivoululumigo:
One big in python 3.12 is that the imp module is deprecated & no longer supported, and our snapshot of the `awslambdaric` code used that module.
In order to solve this I had to add two versions of `awslambdaric` to this repo, the latest version (2.0.8) that support python 3.7 - 3.12, and version 2.0.0 that is the latest one to support python 3.6.
In addition I had to add `simplejson` library to this codebase, because both versions of `awslambdaric` need it.

The source code of these libraries wasn't modified except for imports from `lambdaric` to `simplejson` and auto formatting done by black cli took

src/lumigo_tracer/libs/__init__.py
```diff
@@ -1 +1,11 @@
 # TODO - do we need to add to our license something that thanks those repositories?
```
saartochner, line 1: remove the `TODO`
Replace "customers" with "users"

src/lumigo_tracer/libs/__init__.py
```diff
@@ -1 +1,11 @@
...
+
+"""
+Lumigo Tracer tries to use as little external libraries as possible, but sometimes it's just not possible.
+in these cases (When the library is small enough) we copy the code to our repository and use it from there, in order to
+avoid dependency version conflict with the customers dependencies.
+
+The source code is not altered in any way, except for:
+* Imports between libraries are changed to use the local copy.
+* Auto formatting is applied.
```
saartochner, line 10: why? we should ignore these paths in black...

src/lumigo_tracer/libs/awslambdaric/release_2_0_0/runtime_client.cpp
```diff
@@ -0,0 +1,148 @@
+/* Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved. */
```
saartochner, line 1: I don't think that we should have code that we don't use. I would trim all these new files only to the bare necessary

src/lumigo_tracer/auto_instrument_handler.py
```diff
@@ -1,13 +1,25 @@
...
-    # Import a snapshot of _get_handler from python38 runtime
-    from lumigo_tracer.libs.bootstrap import _get_handler as aws_get_handler
+    # Import a snapshot of _get_handler from the lambda runtime interface client (awslambdaric)
+    # matching the current python version
+    if sys.version_info >= (3, 7):
+        # This version supports python 3.7 up to 3.12
+        from lumigo_tracer.libs.awslambdaric.release_2_0_8.bootstrap import (
+            _get_handler as aws_get_handler,
+        )
+    else:
```
saartochner, line 17: are there cases that are supported in python 3.6 and not >3.7.?
Maybe the first import is sufficient?

In any case, I would remove it because you know for sure that in python3.6, you'll not get into this "except" - because the "try" will succeed for sure.
Initially, we created this fallback because we feared the unknown. Your case is really not unknown :)

src/test/unit/test_auto_instrument_handler.py
```diff
@@ -1,11 +1,23 @@
...
+import sys
 import traceback
 
 import mock
 import pytest
 
 from lumigo_tracer.auto_instrument_handler import ORIGINAL_HANDLER_KEY, _handler
-from lumigo_tracer.libs.lambda_runtime_exception import FaultException
+
+if sys.version_info >= (3, 7):
```
saartochner, line 10: AFAIK, the unit tests run once during our CI: with python3.7. 
I.e. your `else` will never be executed as part of our CI, so this is basically a stale code.
IMO - you don't need it at all. 3.6 is EOF

src/lumigo_tracer/libs/__init__.py
sagivoululumigo: I am waiting for an answer from David to see if we have any limitations / requirements because we are copying source code from external libraries

src/lumigo_tracer/libs/__init__.py
sagivoululumigo: Correct, ignored them

src/test/unit/test_auto_instrument_handler.py
sagivoululumigo: I wanted to deprecate support for 3.6, but it turns out that we still have active customers using lambda 3.6...

Anyways because the imported library is for worst case scenario, I feel good only using the awslambdaric version that supports 3.7+ (It unofficially supports 3.6 as specified in their setup.py file in version 2.0.8)

src/lumigo_tracer/auto_instrument_handler.py
sagivoululumigo: awslambdaric version 2.0.8 supports (officially) python 3.7 - 3.12 (And unofficially supports 3.6 according to the setup.py file)

awslambdaric version 2.0.0 supports 3.6 - 3.11.

So because the lib is used in a worst case scenario that awslamdbaric isn't available I think we should just use version 2.0.8

src/lumigo_tracer/libs/awslambdaric/release_2_0_0/runtime_client.cpp
sagivoululumigo: Removed

src/lumigo_tracer/auto_instrument_handler.py
```diff
@@ -6,8 +6,19 @@
     # Try to import AWS's current _get_handler logic
     from bootstrap import _get_handler as aws_get_handler
 except Exception:
-    # Import a snapshot of _get_handler from python38 runtime
-    from lumigo_tracer.libs.bootstrap import _get_handler as aws_get_handler
+    # Only import the logger if we need it, no need to slow down the import time
+    from lumigo_core.logger import get_logger
```
sagivoululumigo, line 10: I only import the logger if we need to log a warning because I don't want to complicate this basic file with extra imports & logic if we don't need to

saartochner: there is `warn_client`, which I think is what you want

src/lumigo_tracer/libs/awslambdaric/bootstrap.py
```diff
@@ -2,44 +2,30 @@
...
         (modname, fname) = handler.rsplit(".", 1)
     except ValueError as e:
         fault = FaultException(
-            FaultException.MALFORMED_HANDLER_NAME, "Bad handler '{}': {}".format(handler, str(e))
+            FaultException.MALFORMED_HANDLER_NAME,
+            "Bad handler '{}': {}".format(handler, str(e)),
         )
         return make_fault_handler(fault)
 
-    file_handle, pathname, desc = None, None, None
```
saartochner, line 26: just making sure: this is all copied from the latest version, correct?

sagivoululumigo: Yes, awslambdaric 2.0.8

PR #301, opened 2024-01-08
feat: Rd 11770 dynamic payload capture & smart spans prioritization

nadav3396:


src/lumigo_tracer/lambda_tracer/lambda_reporter.py
```diff
@@ -168,36 +182,135 @@ def report_json(
     return duration
 
 
+def get_span_priority(span: Dict[Any, Any]) -> int:
+    if span.get("type") == FUNCTION_TYPE:
+        return 0
+    if is_span_has_error(span):
+        return 1
+    if span.get("type") == ENRICHMENT_TYPE:
+        return 2
```
saartochner, line 191: IMO encrichment is more important than error - there is up to one enrichment, it is super small, and contain metadata on the invocation (execution tags).
This is probably more important than one more error span...

src/lumigo_tracer/lambda_tracer/lambda_reporter.py
```diff
@@ -168,36 +182,135 @@ def report_json(
...
+            span_copy.pop("response", None)
+            return span_copy
+        if span_type == SQL_SPAN:
+            span_copy.pop("query", None)
+            span_copy.pop("values", None)
+            span_copy.pop("response", None)
+            return span_copy
+
+    get_logger().warning(f"Got unsupported span type: {span_type}")
+    return span
```
saartochner, line 226: the fallback transmit the whole span as is? maybe return an empty dict?

src/lumigo_tracer/lambda_tracer/lambda_reporter.py
```diff
@@ -168,36 +182,135 @@ def report_json(
...
+                if span_metadata == {}:
+                    continue
+                span_metadata_size = _get_event_base64_size(span_metadata)
+
+                if span.get("type") == FUNCTION_TYPE:
+                    # We always want to at least send the function span
+                    spans_to_send_dict[index] = span_metadata
+                    spans_to_send_sizes[index] = span_metadata_size
+                    current_size += span_metadata_size
+                elif current_size + span_metadata_size < max_size:
```
saartochner, line 287:
```suggestion
                if span.get("type") == FUNCTION_TYPE or current_size + span_metadata_size < max_size:
                    # We always want to at least send the function span
                    spans_to_send_dict[index] = span_metadata
                    spans_to_send_sizes[index] = span_metadata_size
                    current_size += span_metadata_size
```
right? did i missread?

src/lumigo_tracer/lambda_tracer/lambda_reporter.py
```diff
@@ -168,36 +182,135 @@ def report_json(
...
-                break
-    return aws_dump(spans_to_send)[:max_size]
+        if current_size + span_size > max_size:
+            break
+
+        spans_to_send.append(span)
+        current_size += span_size
+
+    if len(spans_to_send) < len(msgs):
+        with lumigo_safe_execute("create_request_body: smart span selection"):
```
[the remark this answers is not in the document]
saartochner: and give it a nice section of doc. this is not a trivial logic

src/lumigo_tracer/lambda_tracer/lambda_reporter.py
nadav3396: I agree I will take this with Or and decide :)

src/lumigo_tracer/lambda_tracer/lambda_reporter.py
```diff
@@ -168,36 +182,135 @@ def report_json(
     return duration
 
 
+def get_span_priority(span: Dict[Any, Any]) -> int:
+    if span.get("type") == FUNCTION_TYPE:
+        return 0
+    if is_span_has_error(span):
+        return 1
```
saartochner, line 189: It will be hard to explain this to the user.
I'm suggesting a second ordering - by time!
first include error spans, but if there are too much - include the chronological first of these..

src/lumigo_tracer/lambda_tracer/lambda_reporter.py
```diff
@@ -168,36 +182,135 @@ def report_json(
...
-            if too_big_spans == too_big_spans_threshold:
-                break
-    return aws_dump(spans_to_send)[:max_size]
+        if current_size + span_size > max_size:
+            break
+
+        spans_to_send.append(span)
+        current_size += span_size
+
+    if len(spans_to_send) < len(msgs):
```
saartochner, line 314: note that you're changing the current behavior of the tracer.
To be fair for our users, allow them to skip this update with a configuring env var. Something like "LUMIGO_SPAN_METADATA_UPON_LIMIT" which is True by default

src/lumigo_tracer/lambda_tracer/spans_container.py
```diff
@@ -246,7 +245,7 @@ def add_exception_to_span(  # type: ignore[no-untyped-def]
             message = str(message)
         span["error"] = SpansContainer._create_exception_event(
             exc_type=exception.__class__.__name__,
-            message=message,
+            message=lumigo_dumps(message),
```
saartochner, line 248: message is string. Do you double escape here?

src/lumigo_tracer/lambda_tracer/lambda_reporter.py
```diff
@@ -37,13 +38,26 @@
...
 NUMBER_OF_SPANS_IN_REPORT_OPTIMIZATION = 200
 COOLDOWN_AFTER_TIMEOUT_DURATION = datetime.timedelta(seconds=10)
 CHINA_REGION = "cn-northwest-1"
 LUMIGO_SPANS_DIR = "/tmp/lumigo-spans"
+FUNCTION_TYPE = "function"
+ENRICHMENT_TYPE = "enrichment"
+HTTP_TYPE = "http"
+MONGO_SPAN = "mongoDb"
+REDIS_SPAN = "redis"
+SQL_SPAN = "mySql"
```
saartochner, line 59: this is a great change! 🙏

src/test/unit/wrappers/__init__.py
saartochner, line 1: why did you add al those init files?

src/test/unit/wrappers/http/test_sync_http_wrappers.py
```diff
@@ -316,7 +316,7 @@ def lambda_test_function(event, context):
 
     assert len(SpansContainer.get_span().spans) == 1
     span = list(SpansContainer.get_span().spans.values())[0]
-    assert span["error"]["message"] == "division by zero"
+    assert span["error"]["message"] == '"division by zero"'
```
saartochner, line 319: ohhhhh I told you that you've double escaped! not good...

src/test/unit/lambda_tracer/test_lambda_reporter.py
```diff
@@ -28,42 +35,190 @@
...
     assert _create_request_body([dummy_span], True, 1_000_000) == json.dumps([dummy_span])
 
 
 def test_create_request_body_keep_function_span_and_filter_other_spans(
-    dummy_span, function_end_span
+    dummy_span: dict, function_end_span: dict
 ):
-    expected_result = [dummy_span, dummy_span, dummy_span, function_end_span]
+    input_spans = [dummy_span, dummy_span, dummy_span, function_end_span, function_end_span]
+    expected_result = [function_end_span, function_end_span, dummy_span]
```
saartochner, line 208: where are all the dummy_spans? why is there only one left?

src/test/unit/lambda_tracer/test_lambda_reporter.py
```diff
@@ -73,7 +228,67 @@ def test_create_request_body_take_error_first(dummy_span, error_span, function_e
...
+def test_create_request_body_take_only_metadata_function_span(
+    function_end_span: dict, function_end_span_metadata: dict
+):
+    expected_result = [function_end_span_metadata]
+    input_spans = [function_end_span]
+    size = _get_event_base64_size(expected_result)
+
+    result = _create_request_body(input_spans, True, max_size=size, max_error_size=size)
+
+    assert result == json.dumps(expected_result)
```
saartochner, line 245: can you check the log?
If there's no log - then add one! this will make our life much easier when debugging

src/lumigo_tracer/lambda_tracer/spans_container.py
nadav3396: I want to limit its size

src/test/unit/wrappers/__init__.py
nadav3396: This is the right way, no? every python directory should have init file

src/test/unit/lambda_tracer/test_lambda_reporter.py
```diff
@@ -28,42 +35,190 @@
...
-    expected_result = [dummy_span, dummy_span, dummy_span, function_end_span]
+    input_spans = [dummy_span, dummy_span, dummy_span, function_end_span, function_end_span]
+    expected_result = [function_end_span, function_end_span, dummy_span]
     size = _get_event_base64_size(expected_result)
-    assert _create_request_body(expected_result * 2, True, size) == json.dumps(
-        [function_end_span, dummy_span, dummy_span, dummy_span]
-    )
 
+    result = _create_request_body(input_spans, True, size)
+    print(result)
```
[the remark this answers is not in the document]
nadav3396: ooooopssss

src/test/unit/lambda_tracer/test_lambda_reporter.py
```diff
@@ -73,7 +228,67 @@ def test_create_request_body_take_error_first(dummy_span, error_span, function_e
...
+def test_create_request_body_take_only_metadata_pymongo_span(
+    function_end_span: dict, pymongo_span: dict, pymongo_span_metadata: dict
+):
+    assert_use_metadata_span_when_needed(function_end_span, pymongo_span, pymongo_span_metadata)
+
+
+def test_create_request_body_take_only_metadata_sql_span(
+    function_end_span: dict, sql_span: dict, sql_span_metadata: dict
+):
+    assert_use_metadata_span_when_needed(function_end_span, sql_span, sql_span_metadata)
```
saartochner, line 281: don't you think that this is all a single test with parameterized? 🤔

src/test/unit/lambda_tracer/test_lambda_reporter.py
```diff
@@ -73,7 +228,67 @@ def test_create_request_body_take_error_first(dummy_span, error_span, function_e
...
+def assert_use_metadata_span_when_needed(
+    function_span, wrapper_span, wrapper_span_metadata
+) -> None:
+    expected_result = [function_span, wrapper_span_metadata]
+    input_spans = [wrapper_span, function_span]
+    size = _get_event_base64_size(expected_result)
+
+    result = _create_request_body(input_spans, True, max_size=size, max_error_size=size)
+
+    assert result == json.dumps(expected_result)
```
saartochner, line 257: I would add a check to the log here too

src/test/unit/lambda_tracer/test_lambda_reporter.py
```diff
@@ -73,7 +228,67 @@ def test_create_request_body_take_error_first(dummy_span, error_span, function_e
...
+def test_create_request_body_take_only_metadata_sql_span(
+    function_end_span: dict, sql_span: dict, sql_span_metadata: dict
+):
+    assert_use_metadata_span_when_needed(function_end_span, sql_span, sql_span_metadata)
+
+
+def test_with_many_spans(function_end_span, http_span, http_span_metadata):
+    expected_result = [function_end_span] + [http_span] * 51 + [http_span_metadata] * 49
+    input_spans = [function_end_span] + [http_span] * 100
+    size = 78080
```
saartochner, line 287: ohhh this test will break the next time that someone change something small (e.g. the size HTTP metadata, a new field to HTTP span, etc.)
Can you make it more robust?

saartochner (changes requested):
GJ!

src/test/unit/lambda_tracer/test_lambda_reporter.py
nadav3396: yes but I can't  parametrize fixtures, no?

src/test/unit/lambda_tracer/test_lambda_reporter.py
```diff
@@ -28,42 +35,190 @@
...
+        "requestCommand": '["SET", "GET"]',
+        "token": "t_token",
+        "ended": now.timestamp() * 1000,
+        "response": '"Result"',
+    }
+
+
+@fixture
+def redis_span_metadata(redis_span: dict) -> dict:
+    span_copy = copy.deepcopy(redis_span)
```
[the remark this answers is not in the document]
nadav3396: But I want to check `the get_span_metadata` logic.

src/test/unit/lambda_tracer/test_lambda_reporter.py
```diff
@@ -28,42 +35,190 @@
...
+        "parentId": "1234",
+        "lambda_container_id": "4062b9eb-5f2d-4dde-9983-3f4404f30b5a",
+        "token": "t_10faa5e13e7844aaa1234",
+        "ended": now.timestamp() * 1000,
+    }
+
+
+@fixture
+def http_span_metadata(http_span: dict) -> dict:
+    span_copy = copy.deepcopy(http_span)
```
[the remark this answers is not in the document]
nadav3396: same

src/test/unit/lambda_tracer/test_lambda_reporter.py
nadav3396: because the size us limited, I wanted to check that we prioritize the `end_span` before other spans.

src/lumigo_tracer/lambda_tracer/lambda_reporter.py
nadav3396: Im not sure I understand, you want to add more filed to sort by no?
how its easier to explain?

src/lumigo_tracer/lambda_tracer/lambda_reporter.py
nadav3396: Or agreed

src/lumigo_tracer/lambda_tracer/lambda_reporter.py
nadav3396: I talked with Orr and he don't want to enable the user to do this.
So Im leaving it like this for now.

src/lumigo_tracer/lambda_tracer/lambda_reporter.py
saartochner: exmple: 
assume an invocation with 1000 successful HTTP requests.
I think that it makes sense that Lumigo will present the first 200, and not random 200 (as currently happens)

src/test/unit/wrappers/__init__.py
saartochner: not tests
init file means that this is a module. unless you do some import (e.g. for test-utils or something similar) then this is not necessary...

src/test/unit/lambda_tracer/test_lambda_reporter.py
saartochner: I'm arguing that you don't need fixtures here

src/lumigo_tracer/lambda_tracer/lambda_reporter.py
saartochner: that's surprising..
ok then...

src/test/unit/lambda_tracer/test_lambda_reporter.py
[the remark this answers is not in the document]
saartochner: do you want to test `get_span_metadata` logic or `_create_request_body`?
the only place that uses this logic executes `_create_request_body`, and not `get_span_metadata`.

If your goal is to test `get_span_metadata`, then call it directly. 
Otherwise, you only want to test that `get_span_metadata` is being called internally - i.e. you should use it to create the expected outcome.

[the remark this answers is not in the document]
saartochner: I'll clearify that I think that all this one-time-used fixtures should be deleted, and you should use `parameterized` instead of a single-lined-test that calls a util :)

saartochner (approved):

src/lumigo_tracer/lambda_tracer/lambda_reporter.py
nadav3396: I understand that, but like I said your are suggestion something more to sort by so by definition its something more we need to explain => meaning it will be harder, no?

we can do it but I don't think its so important.

src/test/unit/wrappers/__init__.py
nadav3396: You are right its not necessary but from what Im familiar its a convention (from a fast check I did all the open sources are like this isort, poetry, etc...) but to be fair pytest themself don't do this but I think I will leave it like this :)

PR #313, opened 2024-08-20
feat: report dropped span count

sagivoululumigo:


src/lumigo_tracer/lambda_tracer/lambda_reporter.py
```diff
@@ -275,14 +288,51 @@ def _get_prioritized_spans(
...
+                    spans_to_send_dict[enrichment_spans_index]
+                )
+                size_increase = enrichment_span_size_after - enrichment_span_size_before
+                current_size += size_increase
+                if current_size > buffered_max_size:
+                    # TODO: Remove more spans based on the order until we are within the limit
+                    get_logger().warning(
+                        f"Enrichment span size increased by {size_increase} bytes, "
+                        f"making the total size too big: {current_size} bytes (max: {buffered_max_size} bytes)"
+                    )
```
eyalgolan, line 334: Can we have this in a separate method? something like `update_enrichment_spans`? The `_get_prioritized_spans` is already quite big and getting a little hard to follow what's going on.

src/lumigo_tracer/lambda_tracer/lambda_reporter.py
```diff
@@ -46,6 +48,9 @@
     int(os.environ.get("LUMIGO_MAX_SIZE_FOR_REQUEST_ON_ERROR", 1024 * 990)), REQUEST_MAX_SIZE
 )
 MAX_NUMBER_OF_SPANS: int = int(os.environ.get("LUMIGO_MAX_NUMBER_OF_SPANS", 2000))
+
+# Size of spans sent that is kept for the enrichment span additional info added during sending
+SPANS_SEND_SIZE_ENRICHMENT_SPAN_BUFFER = 200
```
eyalgolan, line 53: Why 200?

src/lumigo_tracer/lambda_tracer/lambda_reporter.py
sagivoululumigo: Done

src/lumigo_tracer/lambda_tracer/lambda_reporter.py
sagivoululumigo: 200 is a const large enough to add
```
dropped_spans_reasons[DroppedSpansReasons.SPANS_SENT_SIZE_LIMIT.value] = {
            "drops": len(msgs) - len(spans_dict),
        }
```

I don't think we need to over complicate it with a dynamic buffer or something like that

src/lumigo_tracer/lambda_tracer/lambda_reporter.py
```diff
@@ -275,17 +289,77 @@ def _get_prioritized_spans(
...
                 spans_to_send_dict[index] = span
                 current_size += span_size - span_metadata_size
             else:
                 # This is an optimization step. If the spans are too big, don't try to send them.
                 too_big_spans += 1
                 if too_big_spans >= too_big_spans_threshold:
                     break
+
+            # If we dropped spans we need to update the enrichment spans dropped spans reasons
+            if len(spans_to_send_dict) != len(msgs):
```
nadav3396, line 302: don't put the condition here, it need to be outside the for loop, here the break can skip it

src/lumigo_tracer/lambda_tracer/lambda_reporter.py
```diff
@@ -275,17 +289,77 @@ def _get_prioritized_spans(
...
+    """
+    Looks at the given spans about to be sent + the total number of messages,
+    and updates the enrichment spans about any dropped spans
+    @param spans_dict: The mapping of spans about to be sent
+    @param msgs: The complete list of spans created
+    @param current_size: The current size of all the spans in spans_dict
+    @param max_size: The maximum size of all spans together
+    @return: A modified version of spans_dict input, including the updated enrichment span
+    """
+    spans_to_send_dict = {index: dict(span) for index, span in spans_dict.items()}
```
nadav3396, line 328: I think this is a waste of time to convert all spans and iterate them twice, mostly that the enrichment span should be the second spans, its better to just do a for loop that run the logic when it see the first enrichment span and add the data to it.

src/lumigo_tracer/lambda_tracer/lambda_reporter.py
```diff
@@ -275,17 +289,77 @@ def _get_prioritized_spans(
...
+        spans_to_send_dict[enrichment_spans_index][
+            DROPPED_SPANS_REASONS_KEY
+        ] = dropped_spans_reasons
+        enrichment_span_size_after = get_event_base64_size(
+            spans_to_send_dict[enrichment_spans_index]
+        )
+        size_increase = enrichment_span_size_after - enrichment_span_size_before
+        updated_size = current_size + size_increase
+        if updated_size > max_size:
+            get_logger().warning(
```
nadav3396, line 355: thats not good, you cant just write a warning in this case, if the request is bigger than 1MB the data will be lost.

src/lumigo_tracer/lambda_tracer/lambda_reporter.py
```diff
@@ -275,17 +289,77 @@ def _get_prioritized_spans(
...
+                    "create_request_body: smart span selection: updating enrichment span"
+                ):
+                    spans_to_send_dict = _update_enrichment_span_about_prioritized_spans(
+                        spans_to_send_dict, msgs, current_size, buffered_max_size
+                    )
+
     return list(spans_to_send_dict.values())
 
 
+def _update_enrichment_span_about_prioritized_spans(
```
nadav3396, line 313: I think you should try to rewrite this method the goals should be
1. avoid iterating the spans many times
2. try to not update dictionaries inside span
3. make sure we cant cross the limit size

my suggestion is to not add the enrichment span to the list but keep space for it and just add it once in the end.

nadav3396 (changes requested):
look on the comments

src/lumigo_tracer/lambda_tracer/lambda_reporter.py
sagivoululumigo: Done

src/lumigo_tracer/lambda_tracer/lambda_reporter.py
sagivoululumigo: Fixed, now if we cross the size limit we log an error and return the original list of spans

src/lumigo_tracer/lambda_tracer/lambda_reporter.py
sagivoululumigo: Fixed in the refactor of the func

src/lumigo_tracer/lambda_tracer/lambda_reporter.py
sagivoululumigo: Added it to the loop by mistake, fixed

src/lumigo_tracer/lambda_tracer/lambda_reporter.py
```diff
@@ -275,15 +289,84 @@ def _get_prioritized_spans(
...
+    spans_size = current_size
+    for span in spans_dict.values():
+        if span.get("type") == ENRICHMENT_TYPE:
+            enrichment_spans.append(span)
+            current_span_size = get_event_base64_size(span)
+            spans_size -= current_span_size
+        else:
+            spans.append(span)
+
+    if not not enrichment_spans or len(enrichment_spans) > 1:
```
nadav3396, line 341: not not

src/lumigo_tracer/lambda_tracer/lambda_reporter.py
```diff
@@ -275,15 +289,84 @@ def _get_prioritized_spans(
...
                 current_size += span_size - span_metadata_size
             else:
                 # This is an optimization step. If the spans are too big, don't try to send them.
                 too_big_spans += 1
                 if too_big_spans >= too_big_spans_threshold:
                     break
-    return list(spans_to_send_dict.values())
+
+        # If we dropped spans we need to update the enrichment spans dropped spans reasons
+        final_spans_list = list(spans_to_send_dict.values())
```
nadav3396, line 302: put this line in the else of them if

src/lumigo_tracer/lambda_tracer/lambda_reporter.py
```diff
@@ -275,15 +289,84 @@ def _get_prioritized_spans(
...
+        else:
+            spans.append(span)
+
+    if not not enrichment_spans or len(enrichment_spans) > 1:
+        # We should never get here, if we did it probably means a bug in the tracer code.
+        get_logger().warning(f"Got unsupported number of enrichment spans: {len(enrichment_spans)}")
+        return list(spans_dict.values())
+
+    dropped_spans_due_to_size_limit = len(msgs) - len(spans_dict)
+    if not dropped_spans_due_to_size_limit:
```
nadav3396, line 347: I think its more clear to write `dropped_spans_due_to_size_limit == 0`

src/lumigo_tracer/lambda_tracer/lambda_reporter.py
```diff
@@ -275,15 +289,84 @@ def _get_prioritized_spans(
...
+                "create_request_body: smart span selection: updating enrichment span"
+            ):
+                final_spans_list = _update_enrichment_span_about_prioritized_spans(
+                    spans_to_send_dict, msgs, current_size, request_max_size
+                )
+
+    return final_spans_list
+
+
+def _update_enrichment_span_about_prioritized_spans(
```
nadav3396, line 314: im not sure I saw where do you test this?

src/lumigo_tracer/lambda_tracer/lambda_reporter.py
sagivoululumigo: I didn't do that on purpose, in case the `lumigo_safe_execute` crashes I still want an answer to return (Better return the spans without returning drops then not returning any data)

src/lumigo_tracer/lambda_tracer/lambda_reporter.py
sagivoululumigo: Done

src/lumigo_tracer/lambda_tracer/lambda_reporter.py
sagivoululumigo: Fixed

src/lumigo_tracer/lambda_tracer/lambda_reporter.py
sagivoululumigo: Added the missing UT

nadav3396 (approved):

